PR #2, opened 2020-06-03
[small] removed references to mern

chriskrogh:


src/db/config.ts
```diff
@@ -1,3 +1,3 @@
 export const port = '27017';
 export const host = '127.0.0.1';
-export const dbName = 'mern';
\ No newline at end of file
+export const dbName = 'acs';
```
jtaaa, line 3: Missing a newline at the end of this file. Weird that eslint didn't catch this on save.
```suggestion
export const dbName = 'acs';

```

chriskrogh: Yea I we should add that to our eslint config. I don't think I have it

PR #5, opened 2020-06-10
[infra] User Context

chriskrogh:
## What's changed

After user authenticates, we want to store their info as well as their auth token (generated by the server) in context so that it's accessible throughout the client app.

This PR introduces actions, a reducer and a provider that allow this to happen.

### Types 

I also realized that the user type needed a slight tweak. 
After the user authenticates, we don't want to keep track of their secret for security reasons.

(A user's secret is their password or something private if they use a 3p auth service like Facebook / Google)

Any time the server returns a user it deletes the secret KVP from the doc (so secret should be optional on a `User`).

However we do want the `Credentials` type to have both email and secret.

### DIFF
--- a/client/src/types/user.ts
+++ b/client/src/types/user.ts
@@ -2,12 +2,13 @@ import { ModelMetadata } from './model';
 
 export type Credentials = {
   email: string;
-  secret?: string;
+  secret: string;
 }
 
 export type UserData = Credentials & {
   name: string;
   picture?: string;
+  secret?: string;
 };
 
 export type User = UserData & ModelMetadata;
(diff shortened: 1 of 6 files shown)


client/src/types/user.ts
```diff
@@ -2,12 +2,13 @@ import { ModelMetadata } from './model';
 
 export type Credentials = {
   email: string;
-  secret?: string;
+  secret: string;
 }
 
 export type UserData = Credentials & {
```
jtaaa, line 8: I think `secret` is still going to be a required property on the `UserData` type since it's required on `Credentials`.

`string & (string | undefined)` &rarr; `(string & string) | (string & undefined)` &rarr; `string | never` &rarr; `string` :(

I hate that it doesn't just override it like you might think but I bet there's some good explanation somewhere with something to do with sets or something or the other 👎 

We can do
```suggestion
export type UserData = Omit<Credentials, 'secret'> & {
```

PR #6, opened 2020-06-10
[easy][infra] Cors as dev dep

chriskrogh:
Conditionally importing `cors` and using it when we're not in `production`
`cors` is now a dev dependency


app.ts
```diff
@@ -11,12 +10,17 @@ const app = express();
 let CLIENT_DIR = 'client/build';
 
 if (path.basename(__dirname) === 'build') {
-  CLIENT_DIR = '../' + CLIENT_DIR;
+  CLIENT_DIR = path.join('..', CLIENT_DIR);
```
chriskrogh, line 13: sneaking this 1 liner in

app.ts
```diff
@@ -11,12 +10,17 @@ const app = express();
...
 }
 
 // Serve static files from the React app
 app.use(express.static(path.join(__dirname, CLIENT_DIR)));
-app.use(cors());
+
+if(process.env.NODE_ENV !== 'production') {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  app.use(require('cors'));
```
chriskrogh, line 21: I think conditional imports is one of the few cases where we can disable this `eslint` rule

jtaaa: Fine by me 👍

PR #7, opened 2020-06-10
[infra] Router for pages

chriskrogh:
Added `react-router-dom` and 2 pages

(Probably ignore the `yarn.lock` diffs)

### Sidenote 

If we want to modify the typography that we use throughout the app eventually, we'll do it through the theme
see: https://material-ui.com/customization/typography/


client/src/App.tsx
```diff
@@ -1,16 +1,19 @@
...
 function App() {
   return (
     <ThemeContextProvider>
       <ThemeProvider>
-        <Page>
-          <ThemeToggle />
-        </Page>
+        <BrowserRouter>
+          <Route exact path={ROUTES.HOME} component={Home} />
+          <Route path={ROUTES.MEMBERSHIP} component={Membership} />
```
jtaaa, line 15: Should these page `Route`s be in a `Switch` so that only one can ever be rendered?

PR #10, opened 2020-06-11
[so much code bro] Mailing List UI

chriskrogh:
[Trello](https://trello.com/c/x3xIxcNz)

### Note to reviewers
This PR leaves some things incomplete, but it's so massive that I think it should be reviewed now before any more code is written.

I'll leave comments on (what I think to be) the important parts. Everything else is pretty boring.

## UI
Screen-Recording-2020-06-10-at-1 (1)

## This code won't be a part of the main app
This code is not gonna be here for long. Just temporarily until we start the membership page.

## What's changed

This is just a UI shell of the mailing list page that exec wants up soon.
Calling it a shell because the `Submit` button and the social media buttons don't do anything.

**MailingList**
> Logo

> Slider
>> Email Slide
>> Feedback Slide
>> Socials Slide

> ButtonContainer
>> Back
>> Next

## What's left to be done after this
1. Use refs to make the height of the slider exactly what's needed instead of the height of the largest slide
2. Fix state bug. Right now there's a bug where updating the array of interested events doesn't re-render the checkbox. So when you check an event, the array of interested events get's updated properly, but the checkbox doesn't check until you navigate away and come back to the component

See this vid:
Screen-Recording-2020-06-10-at-1 (2)

3. Wire up the submit and social icon buttons


client/src/components/forms/MailingList/index.tsx
```diff
@@ -1,11 +1,141 @@
...
+import /*type*/ { WithStyles, Theme } from '@material-ui/core/styles';
+import { withStyles, createStyles } from '@material-ui/core/styles';
+import { Button } from '@material-ui/core';
+import { useState } from 'react';
+import NavigateNextIcon from '@material-ui/icons/NavigateNext';
+import NavigateBeforeIcon from '@material-ui/icons/NavigateBefore';
+import EmailSlide from './slides/Email';
+import FeedbackSlide from './slides/Feedback';
+import SocialSlide from './slides/Social';
+import { EVENT } from './types';
```
chriskrogh, line 11: Storing KVP's of the events in an enum. Waiting for Danicia to give the list

client/src/components/forms/MailingList/slides/Email.tsx
```diff
@@ -0,0 +1,54 @@
...
+interface Props extends WithStyles<typeof styles> {
+  email: string;
+  setEmail: Dispatch<SetStateAction<string>>;
+}
+
+function Email({ classes, email, setEmail }: Props) {
+  return (
+    <div className={classes.container}>
+      <div className={classes.textContainer}>
+        <Typography variant="h4" className={classes.text}>
```
chriskrogh, line 35: Typography example

client/src/components/forms/MailingList/slides/Feedback.tsx
```diff
@@ -0,0 +1,92 @@
...
+        </Typography>
+      </div>
+      <div className={classes.eventsContainer}>
+        {Object.values(EVENT).map(event => (
+          <FormGroup key={event}>
+            <FormControlLabel
+              control={
+                <Checkbox
+                  size='small'
+                  checked={events.indexOf(event) > -1}
```
chriskrogh, line 63: This is the boy right here. Have to find a different way to check this.

client/src/components/forms/MailingList/types.ts
```diff
@@ -0,0 +1,5 @@
+export enum EVENT {
```
chriskrogh, line 1: When Danicia gives us the list we'll populate this with actual events

client/src/components/forms/MailingList/index.tsx
```diff
@@ -1,11 +1,141 @@
...
+      >
+        {!isFirstSlide(translateValue) && !isLastSlide(translateValue) && (
+          <Button onClick={goLeft} className={classes.button}>
+            <NavigateBeforeIcon />
+            Back
+          </Button>
+        )}
+        {!isLastSlide(translateValue) && (
+          <Button onClick={goRight} className={classes.button}>
+            {translateValue === (-100 * (NUMBER_OF_SLIDES - 2)) ? 'Submit' : 'Next'}
```
jtaaa, line 132: I would've preferred keeping `currentSlide` in the state since it's more meaningful and then calculating the `translateValue` but this works

client/src/components/forms/MailingList/slides/Email.tsx
jtaaa: ✅ ✅

client/src/components/forms/MailingList/slides/Feedback.tsx
```diff
@@ -0,0 +1,92 @@
...
+    } else {
+      events.push(event);
+    }
+    setEvents(events);
+  }
+
+  return (
+    <div className={classes.container}>
+      <div className={classes.textContainer}>
+        <Typography variant="h4" className={classes.text}>
```
jtaaa, line 49: I think we can save defining a `text` class by doing `color="textPrimary"`.
https://material-ui.com/api/typography/#props

Edit: Nvm now seeing you used it lower down.

client/src/components/forms/MailingList/index.tsx
jtaaa: 👍

client/src/components/forms/MailingList/slides/Feedback.tsx
jtaaa: Weird bug... I would think this would work. And because you aren't unmounting and remounting anything it's super strange that it works when you navigate away and back 👀. Not sure if you have any leads already but maybe it's something to do with the translation not displaying the checkbox checked correctly.

jtaaa:
I'm gonna merge this so I can lint it in my next PR

client/src/components/forms/MailingList/slides/Feedback.tsx
jtaaa: Nevermind what I sent above. It's because you are pushing or splicing and setting the same array object instead of creating a new one to set. React thinks the array hasn't changed since it has the same reference so it doesn't rerender. Clicking the buttons updates another state which causes the rerender.

chriskrogh: Daaamn thats actually sick. It's because the reference didn't change. Thanks Josh!

chriskrogh: I think I might change the state to an object instead because now we have a `title` and a `description` for each event.

(You don't have to care about the description from an api perspective. As long as each event has a unique identifier (title) we should be good with that for now)

client/src/components/forms/MailingList/index.tsx
chriskrogh: Makes sense. I'll probably update

chriskrogh: Actually I can't because the slides go like

`0` `-100` `-200`

If the values were positive, I'd be more inclined to use `currentSlide`

PR #11, opened 2020-06-11
[looks big but small][infra] Add and configure Prettier

jtaaa:
The first commit has the actual addition and setup and the second is running prettier on the whole project.

Let me know if you have any concerns or if you wanna change any [options](https://prettier.io/docs/en/options.html).


.eslintrc.json
```diff
@@ -3,39 +3,26 @@
   "parserOptions": {
     "ecmaVersion": 2020
   },
-  "plugins": [
-    "@typescript-eslint",
-    "react",
-    "react-hooks"
-  ],
```
jtaaa, line 10: These boys are added by the configs we extend below so I removed em.

.eslintrc.json
```diff
@@ -3,39 +3,26 @@
...
-    "space-in-parens": [ 1, "never" ],
-    "eol-last": [ 2, "always" ],
-    "no-trailing-spaces": 2,
-    "max-len": [
-      2,
-      {
-        "code": 95,
-        "tabWidth": 2,
-        "ignoreStrings": true
-      }
```
jtaaa, line 38: Since prettier is handling our style now I removed all the eslint style rules.

The most noticeable difference is the array bracket spacing is removed. Prettier doesn't have an option to allow it. Also, the maximum code length is 80 by default with prettier. I set it to 95 initially but it made some `jsx` elements with many props extend really long instead of wrapping so I left it as 80.

.eslintrc.json
```diff
@@ -3,39 +3,26 @@
...
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      { "argsIgnorePattern": "^_" }
```
jtaaa, line 25: This lets us define function args starting with `_` to allow them to be unused. I think this came in handy in another project so I put it here one time.

client/src/components/Page/index.tsx
```diff
@@ -1,30 +1,27 @@
 import React from 'react';
 import { withStyles, createStyles } from '@material-ui/core/styles';
-import /*type*/ { WithStyles, Theme } from '@material-ui/core/styles';
+import { /*type*/ WithStyles, Theme } from '@material-ui/core/styles';
```
jtaaa, line 3: Unfortunately, prettier doesn't understand this so it tosses the comment in the brackets. I'm thinking it shouldn't be a big problem since we can still search and replace once the feature is available.

PR #14, opened 2020-06-13
[big boy][UI] Mailing list

chriskrogh:
### Desktop

Screen-Recording-2020-06-12-at-9

### Mobile

Screen-Recording-2020-06-12-at-9 (1)

## What's changed
1. Created static events based on Danicia's list
2. Cleaned up and made code a little more robust
> Removed some unnecessary style classes
> Created array for slides
3. Added email validation 

I say we try a deploy of the UI and test em

### DIFF
--- a/client/src/components/forms/MailingList/index.tsx
+++ b/client/src/components/forms/MailingList/index.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import { /*type*/ WithStyles, Theme } from '@material-ui/core/styles';
 import { /*type*/ Event } from 'components/forms/MailingList/events';
+
+import React, { createRef, useEffect } from 'react';
 import { withStyles, createStyles } from '@material-ui/core/styles';
 import { Button } from '@material-ui/core';
 import { useState } from 'react';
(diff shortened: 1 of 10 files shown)


client/src/components/forms/MailingList/index.tsx
```diff
@@ -85,49 +92,28 @@ function MailingListForm({ classes }: WithStyles<typeof styles>) {
...
-              : 'Next'}
-            <NavigateNextIcon />
-          </Button>
-        )}
+        {slideArr.map((s, index) => (
+          <div
+            key={index}
+            className={classes.slide}
+            style={{ transform: `translateX(${translateValue}%)` }}
+          >
```
jtaaa, line 100: I think a better way to reuse this `div` (if that's what you're going for) would be to define a small subcomponent since it's not immediately obvious what's going on here. Just a nitpick tho 😅

chriskrogh: A sub component for the slide or for the slider?
I'm down for a slider sub component
What do you think makes it hard to read?

jtaaa: A slider component would be very fresh but I meant the slide here.
It 100% doesn't need to be a subcomponent... that's probably overkill but I find the map and matching the index to be less readable than plainly having 3 of the same `div` with the slides inside. The map makes me have to think a little bit and check to see where the array is defined and if it is ever updated to be sure of what's going on whereas if it were just laid out it's blatantly obvious.
I assume the reason was to reuse the containing `div`?

chriskrogh: Oh I see what you're saying! Yes it was to reuse the `div`
Yea I can switch it back. There really is no major need for an array.
Will switch back!

PR #15, opened 2020-06-13
[medium] Add /api/mailingList POST route

jtaaa:
Big note is that we are using some dudes fork of [google-spreadsheet](https://github.com/theoephraim/node-google-spreadsheet) for TS support. The fork is kinda sketch and you may have to add an explicit `any` type definition in one place to get it to build 😅. This [PR to DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45347) by the same guy will add legit types. I'll switch over as soon as it's merged.

Here's a link to the [spreadsheet where the responses are added](https://docs.google.com/spreadsheets/d/16PEbh4LFsb1klDX-W6kp47_p9acEKVIWNBz-TTDXcDc/edit#gid=0). You'll need to be logged into the ACS Gmail to view.


src/routers/mailingList/constant.ts
```diff
@@ -0,0 +1,20 @@
+export const EVENT_NAMES = [
```
chriskrogh, line 1: Not sure if we need this...
It's just used to create `DEFAULT_EVENT_RESPONSES` right?

src/routers/mailingList/constant.ts
```diff
@@ -0,0 +1,20 @@
...
+  'Talent Show',
+  'Study Buddies',
+  'Lewwe Talk',
+  'Trivia Night',
+  'Workout Sessions',
+  'Dance Sessions',
+];
+
+// Object with event names as keys and all values set to false
+export const DEFAULT_EVENT_RESPONSES = EVENT_NAMES.reduce<
```
chriskrogh, line 15: What made you decide to initialize it this way instead of 
```
export const DEFAULT_EVENT_RESPONSES = {
  'Games Night': false,
  'Movie Night': false,
...
}
```
Is there a typing benefit to doing it this way?

src/routers/mailingList/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './mailingList';
```
chriskrogh, line 1: I think you can write all the code in this file. I can't see router's getting too for now

src/routers/mailingList/mailingList.ts
```diff
@@ -0,0 +1,39 @@
...
+import GoogleCreds from '../../config/acs-web-5aaaf-768e669e2d39.json';
+
+const doc = new GoogleSpreadsheet(
+  '16PEbh4LFsb1klDX-W6kp47_p9acEKVIWNBz-TTDXcDc',
+);
+
+const router = express.Router();
+
+// add to mailing list with responses of interest in events and other feedback
+router.post('/api/mailingList', async (req: Request, res: Response) => {
```
chriskrogh, line 14: Maybe we can create a `utils/constants.ts` and have
```
export enum ROUTES {
  MAILING_LIST = '/api/mailingList'
}
```
So that when we add more routes, we can do
`router.post(ROUTES.MAILING_LIST, ...`

Not too related to this PR but can we add it anyways?

src/routers/mailingList/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { MailingListRequestBody } from './types';
+
+export const isMailingListRequestBody = (
```
chriskrogh, line 3: I think writing one of these for each route will be way too tedious😅

src/routers/mailingList/mailingList.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (!isMailingListRequestBody(req.body)) {
```
chriskrogh, line 15: I think this is great initiative to check the body!💯

But here's where I kind of want to follow a practice for all routers...
I prefer to have all routers wrapped in a `try catch`. 

Instead what you could have done was something like:
```
router.post('/api/mailingList', async (req: Request, res: Response) => {
  try {
    const { email, interestedEvents, otherFeedback } = req.body;
    if(!email || !interestedEvents || !otherFeedback) throw new Error(M.BAD_BODY);
    ...
  } catch (err) {
    console.log(err);
    res.status(400).send(error.message);
  }
});
```

It's kind of similar to what you did, however, it's a little more readable because the route will only have two `res.send()` expressions. One if the request was successful, and one if the request was not.

You could imagine that for bigger routes there might be multiple points of failure and it's super convenient to just 
`throw new Error('Message saying what went wrong');`

What do you think?

src/routers/mailingList/types.ts
```diff
@@ -0,0 +1,5 @@
+export type MailingListRequestBody = {
+  email: string;
+  interestedEvents: string[];
+  otherFeedback: string;
+};
```
chriskrogh, line 5: What do you think about using the `as` keyword to try coerce each type in the route?
For example, when you extract `interestedEvents`, could you do `interestedEvents as string[]`
And if it fails at runtime (meaning the req body had something other than string[]), the user gets back a 400 because of the `try catch`?

src/routers/mailingList/mailingList.ts
```diff
@@ -0,0 +1,39 @@
...
+  for (const eventName of interestedEvents) {
+    eventsResponses[eventName] = true;
+  }
+  const newRow = {
+    Email: email,
+    'Other Feedback': otherFeedback,
+    ...eventsResponses,
+  };
+
+  await responseSheet.addRow(newRow);
```
chriskrogh, line 35: This is sick¡!¡!¡!

src/routers/mailingList/constant.ts
jtaaa: Nope not really just that when/if the `EVENT_NAMES` array is updated this is automatically updated.

src/routers/mailingList/constant.ts
jtaaa: Yeah we actually don't. Initially I was thinking we might

src/routers/mailingList/constant.ts
jtaaa: Since we don't need the `EVENT_NAMES` array I'll change him over to this simpler object 👍

src/routers/mailingList/index.ts
jtaaa: All the code in `mailingList.ts` or all the files?
I'm in the habit of only using the index to list exports but in this case it will probably never expand to export anything else but the mailing list route so I see no harm in moving that code to the index

src/routers/mailingList/utils.ts
jtaaa: Yeah 😅 I had done a little bit of looking and found this fairly popular package [express-validator](https://github.com/express-validator/express-validator). I decided not to include it  for this route since it's probably something we should consider and agree on. Lemme know what you think of it when you get a chance.

src/routers/mailingList/index.ts
chriskrogh: Yea just `mailingList.ts`

src/routers/mailingList/mailingList.ts
jtaaa: True true true def. Also easier to change routes if we ever need to.
Another thing is we can actually mount the router on `ROUTES.MAILING_LIST` and then use relative routes in the router. Like
```
// mailingList/index.ts
...
mailingListRouter.post('/', ...);
...

// app.ts
...
app.use('/api/mailingList', mailingListRouter);
...
```

I like it because we don't have to repeat the same route over and over just the relative route but it does make it so you have to check where the router is mounted (`app.ts`) to know what the full route is.

src/routers/mailingList/mailingList.ts
jtaaa: That makes a lot of sense! We can also specify the HTTP status code in the error that we throw.

src/routers/mailingList/mailingList.ts
jtaaa: `!i!i!i`!

src/routers/mailingList/mailingList.ts
chriskrogh: YES!
Can you do this and test it as a POC? That sounds like a really cool pattern and I'm down to try it out!

src/routers/mailingList/types.ts
jtaaa: Casting in with `as` in TypeScript is actually all compile time so there is no check or throw at runtime.
[express-validator](https://github.com/express-validator/express-validator#documentation) might be a solution for validation. Not sure how it works with TS but I'm happy to cast once some kind of validation passes.

chriskrogh: Oh no way! That's interesting to know. 
Yea we need to have a good way to validate our routes.
Reading [the docs](https://express-validator.github.io/docs/) now and it looks really good.
Can you try it out for this route using like the "out of the box" set up?

chriskrogh:
I think your initiative to type the body was definitely on the right track (maybe something I'm a bit lenient about but shouldn't be). We do need a good, robust way to validate & sanitize our routes 

[This](https://express-validator.github.io/docs/) looks like a package worth having. 
If you get the time, can you integrate it into this PR?

src/routers/mailingList/mailingList.ts
jtaaa: Done.
One thing is now that we aren't repeating the route paths the ROUTES enum I added is far less useful... perhaps even useless rn but maybe it will be useful l8r?

src/routers/mailingList/mailingList.ts
jtaaa: Yoo I forgot about this but we can define our own error handler middleware and we don't 
even have to catch in the routes 😮 
https://expressjs.com/en/guide/error-handling.html

jtaaa: Or rather we might still have to catch unfortunately but we just have to send along the error with `next(err)` and all routes can share the error handler

src/middleware/validate.ts
```diff
@@ -0,0 +1,14 @@
...
+import { validationResult } from 'express-validator';
+
+// Middleware to automatically send 442 response with errors on failed validation
+const validate = (req: Request, res: Response, next: NextFunction) => {
+  const errors = validationResult(req);
+  if (errors.isEmpty()) {
+    return next();
+  }
+
+  return res.status(422).json({ errors: errors.array() });
```
jtaaa, line 11: We can customize how we want to return validation errors for all routes here

src/routers/api.ts
```diff
@@ -0,0 +1,11 @@
+import express from 'express';
+import { ROUTES } from '../utils/constants';
+import userRouter from './user';
+import mailingListRouter from './mailingList';
+
+const router = express.Router();
+
+router.use(ROUTES.USER, userRouter);
+router.use(ROUTES.MAILING_LIST, mailingListRouter);
```
jtaaa, line 9: Route definitions are less useful because they are only used here rn

src/routers/mailingList/validator.ts
```diff
@@ -0,0 +1,19 @@
...
+        check('otherFeedback').isString(),
+      ];
+    default:
+      throw new Error(
+        `Validator for relative path "${route}" on mailingList router does not exist.`,
+      );
+  }
+};
+
+export default validator;
```
jtaaa, line 19: I did a little bit more than "out of the box". I'm thinking we can have one validator function for each router like this that returns the correct validation checks based on the relative route. In this case we only have one route 😅  but this way it scales and stays clean when we have multiple.

I would love to have the route param type be narrowed to only the routes that are defined instead of `string` but not sure it's worth it. We'd have to do something like:
```
import { MAILING_LIST_ROUTES as ROUTES } from './constants';

router.post(ROUTES.ROOT, validator(ROUTES.ROOT), validate, ...);
```
and we'd have to define all the relative routes for each router. What do you think?

src/middleware/validate.ts
chriskrogh: Dammmmmmn

src/routers/api.ts
chriskrogh: Loving it

src/routers/mailingList/constant.ts
```diff
@@ -0,0 +1,13 @@
+// Object with event names as keys and all values set to false
+export const DEFAULT_EVENT_RESPONSES = {
```
chriskrogh, line 2: 🤙

src/routers/mailingList/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      const newRow = {
+        Email: email,
+        'Other Feedback': otherFeedback,
+        ...eventsResponses,
+      };
+
+      await responseSheet.addRow(newRow);
+      return res.send();
+    } catch (err) {
+      next(err);
```
chriskrogh, line 44: Pretty interesting

src/utils/constants.ts
```diff
@@ -0,0 +1,4 @@
+export enum ROUTES {
```
chriskrogh, line 1: beaut

PR #16, opened 2020-06-13
[smol] Hook up mailing list response submission

jtaaa:
[🔌](https://docs.google.com/spreadsheets/d/16PEbh4LFsb1klDX-W6kp47_p9acEKVIWNBz-TTDXcDc/edit#gid=0)

### DIFF
--- a/client/src/components/forms/MailingList/index.tsx
+++ b/client/src/components/forms/MailingList/index.tsx
@@ -7,6 +7,7 @@ import validator from 'validator';
 import EmailSlide from './slides/Email';
 import FeedbackSlide from './slides/Feedback';
 import SocialSlide from './slides/Social';
+import { mailingListSubmit } from 'utils/data/mailingList';
 
 const LOGO_SIZE = 200;
 const TRANSITION_TIME = '0.5s';
@@ -85,6 +86,19 @@ function MailingListForm({ classes }: WithStyles<typeof styles>) {
     }
   };
 
+  const submit = async () => {
+    try {
+      await mailingListSubmit({
+        email,
+        interestedEvents: events.map((event) => event.title),
+        otherFeedback: feedback,
+      });
+      goRight();
+    } catch (err) {
+      console.error(err);
+    }
+  };
+
   return (
     <div className={classes.container}>
       <div className={classes.logoContainer}>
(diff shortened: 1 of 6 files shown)


client/src/utils/network/errorMessages.ts
```diff
@@ -4,3 +4,5 @@ export const THIRD_PARTY_AUTH = 'Could not authenticate with third party';
 export const FETCH_USER = 'Could not fetch user';
 export const FETCH_USERS = 'Could not fetch all users';
 export const LOGOUT = 'Could not log user out';
+
+export const MAILING_LIST = 'Could not submit mailing list response';
```
chriskrogh, line 8:
```suggestion
export const MAILING_LIST = 'Could not submit mailing list request';
```

client/src/utils/network/endpoints.ts
```diff
@@ -1,3 +1,4 @@
 export enum APIRoutes {
   USER = '/api/user',
+  MAILING_LIST = '/api/mailingList',
```
chriskrogh, line 3: Yeaaaaaaaa
We need something like this for the backend bro

client/src/types/mailingList.ts
```diff
@@ -0,0 +1,5 @@
+export type MailingListResponse = {
```
chriskrogh, line 1:
```suggestion
export type MailingListRequest = {
```

client/src/components/forms/MailingList/index.tsx
```diff
@@ -85,6 +86,19 @@ function MailingListForm({ classes }: WithStyles<typeof styles>) {
...
+        email,
+        interestedEvents: events.map((event) => event.title),
+        otherFeedback: feedback,
+      });
+      goRight();
+    } catch (err) {
+      console.error(err);
+    }
+  };
+
```
chriskrogh, line 101: What a beauty😍

jtaaa:
Darn yeah I actually meant response as in the user's response to the mailing list form but now I see how that is v confusing

client/src/components/forms/MailingList/index.tsx
jtaaa: If for any reason it fails the user can open up their console to see why 😏 
#UX

PR #19, opened 2020-06-17
Mailing list ui

chriskrogh:
## What's changed
Addressed changes outlined on [this feedback doc](https://docs.google.com/document/d/10i0pbmF3vico9c1Su6bbGfdoRMnqp_mi1x586C3YGJ0/edit?usp=sharing)

1. Added sc and snapcode
2. Added sm handles
3. Mode fullstops
4. You can skip joining the mailing list and just give event feedback 
5. Added title for all pages

## UI
Screen-Recording-2020-06-17-at-9 (1)


src/routers/mailingList/validator.ts
```diff
@@ -4,9 +4,8 @@ const validator = (route: string) => {
   switch (route) {
     case '/':
       return [
-        check('email').normalizeEmail().isEmail(),
+        check('email').isString(),
```
jtaaa, line 7: This will allow the email field to be `undefined` or falsy (the empty string in our case) but still normalize and validate if a value is provided. We should probably also update the `MailingListRequestBody` type to reflect the optional email field.

```suggestion
        check('email')
          .optional({ checkFalsy: true })
          .normalizeEmail()
          .isEmail(),
```

src/routers/mailingList/validator.ts
```diff
@@ -4,9 +4,8 @@ const validator = (route: string) => {
...
         check('interestedEvents').isArray(),
-        check('interestedEvents.*').notEmpty(),
```
jtaaa, line 9: Why'd we lose this boy? 😢

client/src/components/forms/MailingList/slides/Social.tsx
```diff
@@ -23,54 +24,131 @@ const styles = (theme: Theme) =>
...
+    },
+    snapCodeWrapper: {
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center',
+    },
+    snapCodeContainer: {
+      height: 400,
+      [theme.breakpoints.down('sm')]: {
+        height: 'unset',
```
jtaaa, line 64: Since this is a square image I think we should be fine only setting and overriding widths and that way we don't have to unset anything.

client/src/components/forms/MailingList/slides/Social.tsx
```diff
@@ -23,54 +24,131 @@ const styles = (theme: Theme) =>
...
       <div className={classes.iconsContainer}>
         <Button
+          className={classes.button}
           onClick={() =>
             window.open('https://www.facebook.com/uWaterlooACS', '_blank')
           }
         >
           <FacebookIcon className={classes.icon} />
+          facebook.com/uWaterlooACS
         </Button>
```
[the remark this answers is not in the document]
webveloperchris: Button adds more styling (hover states and alignment stuff)

src/routers/mailingList/validator.ts
webveloperchris: Oh that was my bad. Why do we have that boy? Can't the array be empty?

client/src/components/forms/MailingList/slides/Social.tsx
webveloperchris: Ah yea we don't need that one

src/routers/mailingList/validator.ts
webveloperchris: Ohhhhhhhhhhh this checking that the elements in the array weren't empty? Okay my b we can have em

client/src/components/forms/MailingList/slides/Social.tsx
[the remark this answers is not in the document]
jtaaa: Ohh makes sense. What about wrapping the button in an `a` tag?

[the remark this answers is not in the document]
webveloperchris: Ohh I see what you're saying. Do you want to use an anchor so we can specify target and all that?
I could try it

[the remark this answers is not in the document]
jtaaa: Mostly just because this is really a link and an anchor is most appropriate. I think using an anchor in these cases is good for accessibility and SEO.

I'm tripping tho because having a `button` in an `a` isn't valid HTML.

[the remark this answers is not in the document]
jtaaa: Oh apparently we can add `href` and `target` properties to the button itself and Material UI will use an anchor element. [docs 💸](https://material-ui.com/api/button/#props)

src/routers/mailingList/validator.ts
chriskrogh: I'm going to leave the type because email will either be empty string or a valid email

PR #21, opened 2020-06-19
[~1 hunnit] Add user membership check API route

jtaaa:
I added this route but I realize we don't have designs for what happens when the user checks their membership. Presumably, the user is told that they are an active member, inactive member or not a member (and maybe whether or not they have paid?) and directed to press a button to go to log in, renew membership or sign up.

For now, I'm just returning the whole user object but we'll definitely want to change that in the future. I added the `watIAMUserId` field to the user model but didn't tackle adding anything to track membership status.

How do you see us tracking membership status?
Do we want to have a history of whether a user was active for previous terms or only the current term?

### DIFF
--- a/src/routers/mailingList/validator.ts
+++ b/src/routers/mailingList/validator.ts
@@ -6,8 +6,8 @@ const validator = (route: string) => {
       return [
         check('email')
           .optional({ checkFalsy: true })
-          .normalizeEmail()
-          .isEmail(),
+          .isEmail()
+          .normalizeEmail(),
         check('interestedEvents').isArray(),
         check('interestedEvents.*').notEmpty(),
         check('otherFeedback').isString(),
--- a/src/routers/user/index.ts
+++ b/src/routers/user/index.ts
@@ -1,9 +1,14 @@
-import express, { Request, Response } from 'express';
-import UserModel, { UserDoc } from '../models/user';
-import { UserRequest } from '../types/network';
-import auth from '../middleware/auth';
-import * as M from '../utils/errorMessages';
-import '../db/mongoose';
+import express, { Request, Response, NextFunction } from 'express';
+import validator from 'validator';
+import createError from 'http-errors';
+import UserModel, { UserDoc } from '../../models/user';
+import { UserRequest } from '../../types/network';
+import auth from '../../middleware/auth';
+import * as M from '../../utils/errorMessages';
+import '../../db/mongoose';
+import routeValidator from './routeValidator';
+import validate from '../../middleware/validate';
+import { MembershipCheckRequestBody } from './types';
 
 const router = express.Router();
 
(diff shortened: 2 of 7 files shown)


src/routers/mailingList/validator.ts
```diff
@@ -6,8 +6,8 @@ const validator = (route: string) => {
       return [
         check('email')
           .optional({ checkFalsy: true })
-          .normalizeEmail()
-          .isEmail(),
+          .isEmail()
+          .normalizeEmail(),
```
jtaaa, line 10: Sneaking this switcharoo in here

chriskrogh: 👟

src/routers/user/index.ts
```diff
@@ -192,4 +197,37 @@ router.delete('/:id', auth, async (req: Request, res: Response) => {
...
+router.post(
+  '/membership-check',
+  routeValidator('/membership-check'),
+  validate,
+  async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const { emailOrWatIAMUserId } = req.body as MembershipCheckRequestBody;
+
+      const isEmail = validator.isEmail(emailOrWatIAMUserId);
+      const conditions: Record<string, string> = {};
```
chriskrogh, line 210: I think we can query for the user with `emailOrWatIAMUserId` not so?

src/routers/user/routeValidator.ts
```diff
@@ -0,0 +1,28 @@
...
+          [
+            check('emailOrWatIAMUserId')
+              .isEmail()
+              .withMessage('Invalid email')
+              .normalizeEmail(),
+            check('emailOrWatIAMUserId')
+              .isAlphanumeric()
+              .withMessage('Invalid WatIAM user id')
+              .trim(),
+          ],
```
chriskrogh, line 17: | love these |

src/routers/user/index.ts
```diff
@@ -192,4 +197,37 @@ router.delete('/:id', auth, async (req: Request, res: Response) => {
...
+      } else {
+        conditions.watIAMUserId = emailOrWatIAMUserId;
+      }
+      const user = await UserModel.findOne(conditions);
+
+      if (user === null) {
+        throw createError(
+          404,
+          `${
+            isEmail ? 'Email' : 'WatIAM user id'
```
chriskrogh, line 222: Can use `validator.isEmail(emailOrWatIAMUserId);` here I think. I don't think this route has to care too much about which of the 2 they sent...

What do u think?

chriskrogh:
Also wrt tracking membership status, I'll make a smol PR and we can discuss there

src/routers/user/index.ts
jtaaa: Not sure I understand. I'm using `emailOrWatIAMUserId` and query by either `email`  or `watIAMUserId` depending on if it's an email or not.

Do you mean like doing an `$or` and querying over both fields at the same time?

src/routers/user/index.ts
jtaaa: Might be misunderstanding. `isEmail` is `validator.isEmail(emailOrWatIAMUserId)` I just put it in a variable since I used it twice.

We could just send back "Email or WatIAM user id provided does not match an existing user".  I don't think it hurts to specify tho and it might make it a lil nicer if we want to display the error in the client.

src/routers/user/index.ts
chriskrogh: My bad I'm explaining horribly.

I meant instead of `const user = await UserModel.findOne(conditions);`

You could do `const user = await UserModel.findOne(emailOrWatIAMUserId);` right?

src/routers/user/index.ts
chriskrogh: Yea I like that check 4 sure I think we should keep em.

This was related to the comment above. What I was trying to say is that we should keep this check, but instead of using `isEmail`, we can use `validator.isEmail(emailOrWatIAMUserId)` because the check to distinguish between email and UserId is more valuable here than when you where calling `findOne()`

But I explained poorly🤣

src/routers/user/index.ts
chriskrogh: So you only have to check whether it was an email when you're responding to the request

chriskrogh: Oh wait nevermind. I'm talking str8 🗑. You have to specify the field🤣
Code looks great bro

src/routers/user/index.ts
chriskrogh: I'm rtrdd

chriskrogh:
I forgot that u need to specify the field name when using `findOne`🤣
Code looks gr8 bro

PR #25, opened 2020-06-19
[small][up 4 discussion] Proposed membership types

chriskrogh:
## What's changed
I think we should have 3 membership types and store the current type on the user. 

If that sounds good, I'll also write a cron job to set all current members' membership status to `EXPIRED` at the end of a term.

What do you think?

### DIFF
--- a/src/models/user.ts
+++ b/src/models/user.ts
@@ -1,7 +1,7 @@
 import mongoose from 'mongoose';
 import bcrypt from 'bcryptjs';
 import jwt from 'jsonwebtoken';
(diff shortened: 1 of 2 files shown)


jtaaa:
To document what we talked about
1. We only need 3 states
2. We don't need `NOT_A_MEMBER`
3. `EXPIRED` means **not a member** but was a member in a previous term so we have their info and they can renew easily
4. Members can be paid or unpaid

Hopefully that looks good¡¡¡?

jtaaa:
Also it would be cool to have the cron job export all the users to a csv or spreadsheet before it resets em so we have a log of members for each term.

PR #26, opened 2020-06-19
Use projection to only return membershipStatus

jtaaa:
## What's changed
`/api/user/membership-check` will only return user's `membershipStatus` (along with id in object) instead of all user fields.

## Notes
Cool PR template

### DIFF
--- a/src/routers/user/index.ts
+++ b/src/routers/user/index.ts
@@ -212,7 +212,7 @@ router.post(
       } else {
         conditions.watIAMUserId = emailOrWatIAMUserId;
       }
-      const user = await UserModel.findOne(conditions);
+      const user = await UserModel.findOne(conditions, 'membershipStatus');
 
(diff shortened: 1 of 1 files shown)


src/routers/user/index.ts
```diff
@@ -206,13 +206,10 @@ router.post(
...
-      const conditions: Record<string, string> = {};
-      if (isEmail) {
-        conditions.email = emailOrWatIAMUserId;
-      } else {
-        conditions.watIAMUserId = emailOrWatIAMUserId;
-      }
-      const user = await UserModel.findOne(conditions);
+      const user = await UserModel.findOne(
+        { [isEmail ? 'email' : 'watIAMUserId']: emailOrWatIAMUserId },
+        'membershipStatus',
```
chriskrogh, line 211: Whaaaaaaaaa. That rel bad

PR #27, opened 2020-06-19
Update /api/user/signup route

jtaaa:
## What's changed
On `/api/user/signup` route:
- Set `membershipStatus` based on `paid` field in request body
- Validate request body

## Notes
We lost the `M.SIGN_UP` error message since I'm sending whatever errors arise to the default error handler. Maybe we can send the error as an object with an error property that would contain the originally thrown error and then we can set the message as `M.SIGN_UP`. Then we have less control to send a specific message instead of the generic one when trying to throw from inside the catch block tho.

What do you think?

### DIFF
--- a/src/routers/user/index.ts
+++ b/src/routers/user/index.ts
@@ -7,21 +7,31 @@ import auth from '../../middleware/auth';
 import * as M from '../../utils/errorMessages';
 import routeValidator from './routeValidator';
 import validate from '../../middleware/validate';
-import { MembershipCheckRequestBody } from './types';
+import { MEMBERSHIP_STATUS } from '../../types/user';
+import { MembershipCheckRequestBody, SignInRequestBody } from './types';
 
 const router = express.Router();
 
 // sign up
-router.post('/signup', async (req: Request, res: Response) => {
-  try {
-    const user = new UserModel(req.body);
-    const token = await user.generateAuthToken();
-    res.status(201).send({ user, token });
-  } catch (err) {
-    console.log(err);
-    res.status(400).send(M.SIGN_UP);
-  }
-});
+router.post(
+  '/signup',
+  routeValidator('/signup'),
+  validate,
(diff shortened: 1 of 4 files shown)


src/utils/customValidators.ts
```diff
@@ -0,0 +1,10 @@
+const MIN_PASSWORD_LENGTH = 8;
+export const isPassword = (password: string) => {
```
chriskrogh, line 2: v n🧊

src/routers/user/index.ts
```diff
@@ -7,21 +7,31 @@ import auth from '../../middleware/auth';
...
-  }
-});
+router.post(
+  '/signup',
+  routeValidator('/signup'),
+  validate,
+  async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const signupData = req.body as SignInRequestBody;
+      const membershipStatus = signupData.paid
```
chriskrogh, line 23: I'm not sure if we'll ever create a paid user...
I was thinking that a new user would have to apply to be a paid member, and then someone from exec would have to approve the req.
Even if the person was a pid member before signing up, how do we validate that on the sign up page?

chriskrogh:
I think we can lose the whole "Error Messages" in constants or wherever that is. I think each route will have its own unique error message. No point in extracting. Will create PR to remove

src/routers/user/index.ts
jtaaa: Josh has some stuff in the project proposal about sign up at booth days which is a bit different from regular sign up and they can pay one time. This is under "Current online member".
Members who sign up at booth days get an email to prompt them to choose a password and create an account on the site. I guess we would record their info and whether or not they paid in another db collection. I'm thinking we'll create paid users in this case once they set their password.

chriskrogh: Ah okay. Sounds like a solid approach

PR #28, opened 2020-06-20
[smol] Add WhiteButton custom button component

jtaaa:
## What's changed
Added the `BWButton` custom button following [this documentation](https://material-ui.com/components/buttons/#customized-buttons).

## UI
WhiteButton

## Notes
~I considered using the theme in this component so that if we switch to the dark theme it would be black with a white outline instead. I got thrown off by the use of a dark button on the event image and thinking that if we toggled the theme the image wouldn't change so the button shouldn't.~

~Now I'm thinking that maybe that button can just be handled differently and this one can use theme.~

The button's colours are dependent on the theme colours `background.default` and `text.primary`. This means the black is the text colour and not exactly black. I'm not sure if I should add another color for opposite of background to the theme. What you think?

Oh and I added support for `.tsx` stories since I wanna write my stories in ts.


jtaaa:
Also if you have a better name than `BWButton` I am all 👂s

PR #29, opened 2020-06-21
Reset membership

chriskrogh:
## What's changed
Added cron job to 
1. store member info in a google sheet
2. reset membership list

Here's how I derived the cron expression: https://crontab.guru/#0_0_1_JAN,MAY,SEP_*
We're running this job once on the 1st of Jan, May, Sept

### Membership resets
We just set everyone's membership in the db to `EXPIRED`. Here's a sc of 2 users' membership status after the script was run

One was paid and the other was unpaid

Screen Shot 2020-06-21 at 10 30 45 AM

### Export members
Adds a new sheet to the Google Spreadsheet Doc for each term (each time the script runs) and then exports members to that sheet.

## Notes
Did some manual tests and the job works okay.

### DIFF
--- a/src/jobs/membership.ts
+++ b/src/jobs/membership.ts
@@ -7,11 +7,11 @@ import GoogleCreds from '../config/acs-web-5aaaf-768e669e2d39.json';
 const getTerm = () => {
   const date = new Date();
   const month = date.getMonth();
-  if (month === 1) {
+  if (month === 0) {
     return 'Fall' + (date.getFullYear() - 1);
-  } else if (month === 5) {
+  } else if (month === 4) {
     return 'Winter' + date.getFullYear();
-  } else if (month === 9) {
+  } else if (month === 8) {
     return 'Spring' + date.getFullYear();
   }
 };
--- a/src/jobs/membership.ts
+++ b/src/jobs/membership.ts
@@ -1,16 +1,51 @@
 import cron from 'node-cron';
 import User from '../models/user';
 import { MEMBERSHIP_STATUS } from '../types/user';
+import {
+  GoogleSpreadsheet,
+  WorksheetBasicProperties,
+} from 'google-spreadsheet';
+import GoogleCreds from '../config/acs-web-5aaaf-768e669e2d39.json';
+
+const getTerm = () => {
+  const date = new Date();
+  const month = date.getMonth();
+  if (month === 1) {
+    return 'Fall' + (date.getFullYear() - 1);
+  } else if (month === 5) {
+    return 'Winter' + date.getFullYear();
+  } else if (month === 9) {
+    return 'Spring' + date.getFullYear();
+  }
+};
+
+const doc = new GoogleSpreadsheet(
+  '1Kn3Fc1vmi2dd9I645BxBQZMXXNvIJi1zEGnBiHuHDtk',
+);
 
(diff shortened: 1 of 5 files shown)


src/jobs/membership.ts
```diff
@@ -0,0 +1,51 @@
...
+  GoogleSpreadsheet,
+  WorksheetBasicProperties,
+} from 'google-spreadsheet';
+import GoogleCreds from '../config/acs-web-5aaaf-768e669e2d39.json';
+
+const getTerm = () => {
+  const date = new Date();
+  const month = date.getMonth();
+  if (month === 1) {
+    return 'Fall' + (date.getFullYear() - 1);
```
jtaaa, line 14: Wouldn't `month === 1` correspond to winter? And `date.getMonth()` is zero-based so 1 is February

src/jobs/membership.ts
```diff
@@ -0,0 +1,51 @@
...
+  '1Kn3Fc1vmi2dd9I645BxBQZMXXNvIJi1zEGnBiHuHDtk',
+);
+
+cron.schedule('0 0 1 JAN,MAY,SEP *', async function () {
+  try {
+    await doc.useServiceAccountAuth(GoogleCreds);
+    await doc.loadInfo();
+    const newSheet = await doc.addSheet({
+      title: getTerm(),
+    } as WorksheetBasicProperties);
```
jtaaa, line 32: How come we need a cast here? btw the PR to definitely typed is approved and I think just waiting for another review by the maintainer of `google-spreadsheet`.

src/jobs/membership.ts
```diff
@@ -0,0 +1,51 @@
...
+        { membershipStatus: MEMBERSHIP_STATUS.UNPAID },
+      ],
+    });
+    members.forEach(async (member) => {
+      const { name, email, watIAMUserId, membershipStatus } = member;
+      const newRow = { name, email, watIAMUserId, membershipStatus };
+      await newSheet.addRow(newRow);
+      Object.assign(member, {
+        membershipStatus: MEMBERSHIP_STATUS.EXPIRED,
+      });
```
jtaaa, line 45: How come we don't just do `member.membershipStatus = MEMBERSHIP_STATUS.EXPIRED` here?

chriskrogh: I **honestly** can't remember why I did this🤣 I'll use `member.membershipStatus = MEMBERSHIP_STATUS.EXPIRED`

src/jobs/membership.ts
chriskrogh: Good news! https://github.com/theoephraim/node-google-spreadsheet/pull/287#issuecomment-648273964
I'll make a PR upgrading the package and see if that resolves this

src/jobs/membership.ts
chriskrogh: I think we run the job at the end of the term. 
1st of Jan runs for Fall
1st of May runs for Winter
1st of Sept runs for Spring
lmk if this is okay or if I should leave a comment in the code

jtaaa: Fine by me. Still concerned about `month === 1` being February tho

chriskrogh: ¡¡¡!!! 

Thanks for the catch!!!

PR #30, opened 2020-06-21
Updated membership routes

chriskrogh:
## What's changed
2 things have actually changed in this PR

1. I added a route for setting the membership status of a user
```
type: PATCH
body: { membershipStatus: MEMBERSHIP_STATUS }
query: { id: ObjectId } // id represents the user's id whose membership status you want to update 
```

2. I migrated the membership status check route from the `routes/user/index.ts` to `routes/user/membership.ts` and made the following slight modifications (which are open 4 discussion)
* route is now `/api/user/membership/check` instead of `/api/user/membership-check` because we have more membership routes
* I changed the type of the request from `POST` to `GET`
I think that would be the better type of request since we're requesting info on a doc (see [this W3 article](https://www.w3schools.com/tags/ref_httpmethods.asp) for more info)
* because the type of request is now a get request I moved the `emailOrWatIAMUserId` field from the body to the query
* I changed the null check for a user from `if(user === null)` to `if(!user)`. There could be a case where user is `undefined` and `user === null` will be false. (better to check for all falsy values)

## Update
Ignore the changes to `src/routers/user/index.ts` during review. they're mostly from prettier
This PR got bigger and might be harder to review, but here's some additional things that I changed:
1. Renamed `UserRequest` to `AuthenticatedRequest` 
This is the request that is passed on after `auth` middleware runs
2. Split the "set membership" route into 2. 
The fist one is accessible by any authenticated user. It sets their membership status to `unpaid`. 
```
type: PATCH
route: api/user/membership/unpaid
```
The second is where an admin can set someone's membership status to anything.
```
type: PATCH
route: api/user/membership
query: id -> id of user who's membership status you want to update
body: membershipStatus -> desired status of user
```
WRT these comments
> For the new route, how do you think we should handle security? Right now anyone could hit the endpoint and update their (and anyone else's) status to paid. I'm thinking we can either:
> 
> 1. have one endpoint that is public that handles renewing the user's own membership (moving themselves from expired to unpaid) and one endpoint that requires "admin" status to hit that can move the user from unpaid to paid or possibly set any field on the user (our current `PATCH /api/user/:id`).
> 2. secure this endpoint so that only admins can move users to paid and regular users can only move themselves from expired to unpaid.
> 
> I think renewing their own membership is the only thing a regular user should be able to do so I prefer 1 to have a specific route to handle that.
> 
> For the changes to `membership-check`:
> 
> 1. I'm completely down for the route path change.
> 2. I originally had it as a GET request but I change it to POST because the email is PII and I think it's best practice to keep that stuff out of the URL. I _think_ for us it's not too critical tho so I'll leave that decision to you.
> 3. Same as 2 above
> 4. According to ts `user` is `UserDoc | null` so just checking for null should be fine. I'm okay with using the shorter `!user` tho.

The idea to split the routes up was a good one. It also gave us a reason to introduce admins which is great.

WRT the PII on the req, in an ideal world we'd write some RSA to encrypt and decrypt the ids on the req query (we can if you want) but I'm not too concerned with this because it would be really hard to intercept the req from client to server as a 3rd party and use the PII since we have same origin reqs. (not an ideal solution but RSA would be a nice 2 have rather than necessary I'd say)

Let me know what you think

### DIFF
--- a/src/middleware/auth.ts
+++ b/src/middleware/auth.ts
@@ -1,5 +1,5 @@
 import { Response, NextFunction } from 'express';
-import { UserRequest } from '../types/network';
+import { AuthenticatedRequest } from '../types/network';
 import jwt from 'jsonwebtoken';
 import UserModel from '../models/user';
 
@@ -7,7 +7,11 @@ type TokenVerification = {
   id: string;
 };
 
-const auth = async (req: UserRequest, res: Response, next: NextFunction) => {
+const auth = async (
+  req: AuthenticatedRequest,
+  res: Response,
+  next: NextFunction,
+) => {
   try {
     const { APP_SECRET } = process.env;
     const authHeader = req.header('Authorization');
--- a/src/routers/user/index.ts
+++ b/src/routers/user/index.ts
@@ -1,6 +1,6 @@
 import express, { Request, Response, NextFunction } from 'express';
 import UserModel, { UserDoc } from '../../models/user';
-import { UserRequest } from '../../types/network';
+import { AuthenticatedRequest } from '../../types/network';
 import auth from '../../middleware/auth';
 import * as M from '../../utils/errorMessages';
 import routeValidator from './routeValidator';
@@ -66,39 +66,47 @@ router.post('/thirdPartyAuth', async (req: Request, res: Response) => {
 });
 
 // logout
-router.post('/logout', auth, async (req: UserRequest, res: Response) => {
-  try {
-    if (req.user == null) {
-      throw new Error();
+router.post(
+  '/logout',
+  auth,
+  async (req: AuthenticatedRequest, res: Response) => {
+    try {
+      if (req.user == null) {
+        throw new Error();
+      }
+      req.user.tokens = req.user.tokens.filter((token) => {
+        return token !== req.token;
+      });
+      await req.user.save();
+      res.send();
+    } catch (err) {
+      console.log(err);
+      res.status(500).send();
     }
-    req.user.tokens = req.user.tokens.filter((token) => {
-      return token !== req.token;
-    });
-    await req.user.save();
-    res.send();
-  } catch (err) {
-    console.log(err);
-    res.status(500).send();
-  }
-});
+  },
+);
 
 // logout all
-router.post('/logoutAll', auth, async (req: UserRequest, res: Response) => {
-  try {
-    if (req.user == null) {
-      throw new Error();
+router.post(
+  '/logoutAll',
+  auth,
+  async (req: AuthenticatedRequest, res: Response) => {
+    try {
+      if (req.user == null) {
+        throw new Error();
+      }
+      req.user.tokens = [];
+      await req.user.save();
+      res.send();
+    } catch (err) {
+      console.log(err);
+      res.status(500).send();
     }
-    req.user.tokens = [];
-    await req.user.save();
-    res.send();
-  } catch (err) {
-    console.log(err);
-    res.status(500).send();
-  }
-});
+  },
+);
 
 // get me
-router.get('/me', auth, async (req: UserRequest, res: Response) => {
+router.get('/me', auth, async (req: AuthenticatedRequest, res: Response) => {
   res.send(req.user);
 });
 
@@ -147,7 +155,7 @@ router.get('/:id/picture', async (req: Request, res: Response) => {
 });
 
 // update me
-router.patch('/me', auth, async (req: UserRequest, res: Response) => {
+router.patch('/me', auth, async (req: AuthenticatedRequest, res: Response) => {
   try {
     const source = req.body as UserDoc;
     if (!req.user) throw new Error();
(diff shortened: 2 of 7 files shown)


jtaaa:
For the new route, how do you think we should handle security? Right now anyone could hit the endpoint and update their (and anyone else's) status to paid. I'm thinking we can either:

1. have one endpoint that is public that handles renewing the user's own membership (moving themselves from expired to unpaid) and one endpoint that requires "admin" status to hit that can move the user from unpaid to paid or possibly set any field on the user (our current `PATCH /api/user/:id`).
2. secure this endpoint so that only admins can move users to paid and regular users can only move themselves from expired to unpaid.

I think renewing their own membership is the only thing a regular user should be able to do so I prefer 1 to have a specific route to handle that.

For the changes to `membership-check`:

1. I'm completely down for the route path change.
2. I originally had it as a GET request but I change it to POST because the email is PII and I think it's best practice to keep that stuff out of the URL. I _think_ for us it's not too critical tho so I'll leave that decision to you.
3. Same as 2 above
4. According to ts `user` is `UserDoc | null` so just checking for null should be fine. I'm okay with using the shorter `!user` tho.

src/routers/user/membership.ts
```diff
@@ -0,0 +1,65 @@
...
+  async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const user = await User.findById(req.query.id);
+      if (!user) {
+        throw createHttpError(404, 'Cannot find user with that id');
+      }
+      const membershipStatus = req.body.membershipStatus;
+      if (!Object.values(MEMBERSHIP_STATUS).includes(membershipStatus)) {
+        throw createHttpError(422, 'invalid membership status');
+      }
```
jtaaa, line 25: We can make this into a custom validator and let `express-validator` handle this

src/routers/user/routeValidator.ts
```diff
@@ -12,15 +12,17 @@ const routeValidator = (route: string) => {
...
       ];
-    case '/membership-check':
+    case '/membership':
+      return [check('membershipStatus').notEmpty(), query('id').notEmpty()];
+    case '/membership/check':
       return [
         oneOf(
           [
-            check('emailOrWatIAMUserId')
+            query('emailOrWatIAMUserId')
```
jtaaa, line 21: `check` will validate from all of the query, body, params and a few other places so it doesn't need to be replaced here unless the goal is to be more specific. Then we should replace `check` with `body` elsewhere.

Actually I think we definitely should replace all the `check`s with the more specific `body` or whichever necessary since I bet validation would incorrectly pass if the parameters to request are sent in the wrong place (expected in body but sent in query).

chriskrogh: I agree, we can add that as a refactor

src/routers/user/membership.ts
chriskrogh: Sure!

src/routers/user/routeValidator.ts
chriskrogh: Created a card for this refactor in the BL : https://trello.com/c/LW8TuoI7

chriskrogh:
> For the new route, how do you think we should handle security? Right now anyone could hit the endpoint and update their (and anyone else's) status to paid. I'm thinking we can either:
> 
> 1. have one endpoint that is public that handles renewing the user's own membership (moving themselves from expired to unpaid) and one endpoint that requires "admin" status to hit that can move the user from unpaid to paid or possibly set any field on the user (our current `PATCH /api/user/:id`).
> 2. secure this endpoint so that only admins can move users to paid and regular users can only move themselves from expired to unpaid.
> 
> I think renewing their own membership is the only thing a regular user should be able to do so I prefer 1 to have a specific route to handle that.
> 
> For the changes to `membership-check`:
> 
> 1. I'm completely down for the route path change.
> 2. I originally had it as a GET request but I change it to POST because the email is PII and I think it's best practice to keep that stuff out of the URL. I _think_ for us it's not too critical tho so I'll leave that decision to you.
> 3. Same as 2 above
> 4. According to ts `user` is `UserDoc | null` so just checking for null should be fine. I'm okay with using the shorter `!user` tho.

I just saw this! These are some good points! I'll make some changes and re-request review when I'm done. Closing 4 now

src/middleware/auth.ts
```diff
@@ -1,13 +1,17 @@
...
+import { AuthenticatedRequest } from '../types/network';
 import jwt from 'jsonwebtoken';
 import UserModel from '../models/user';
 
 type TokenVerification = {
   id: string;
 };
 
-const auth = async (req: UserRequest, res: Response, next: NextFunction) => {
+const auth = async (
```
chriskrogh, line 10: prettier

src/routers/user/index.ts
```diff
@@ -67,39 +68,47 @@ router.post('/thirdPartyAuth', async (req: Request, res: Response) => {
 });
 
 // logout
-router.post('/logout', auth, async (req: UserRequest, res: Response) => {
-  try {
-    if (req.user == null) {
-      throw new Error();
+router.post(
```
chriskrogh, line 71: prettier

src/routers/user/index.ts
```diff
@@ -67,39 +68,47 @@ router.post('/thirdPartyAuth', async (req: Request, res: Response) => {
...
-});
+  },
+);
 
 // logout all
-router.post('/logoutAll', auth, async (req: UserRequest, res: Response) => {
-  try {
-    if (req.user == null) {
-      throw new Error();
+router.post(
```
chriskrogh, line 92: prettier

PR #31, opened 2020-06-22
[kinda big] Header UI

jtaaa:
## What's changed
Added [da Header (trello link)](https://trello.com/c/H8rNJunT/15-header-ui). I improvised a mobile header with a drawer from material UI.

## UI (if UI has changed)
Desktop:
HeaderMenu
Mobile (jeez this gif is so tall compared to the other one but I don't wanna take it over 😅 sry):
MobileMenu

## Notes
We can probably improve the mobile menu with Icons for each link and maybe our logo in the drawer. Maybe we can get some feedback from Josh.


.gitignore
```diff
@@ -11,3 +11,5 @@ client/build
 client/node_modules
 
 acs-web-5aaaf-768e669e2d39.json
+
+.DS_Store
```
jtaaa, line 15: f*** this guy

client/src/components/Header/constants.ts
```diff
@@ -0,0 +1,20 @@
+import { MenuLink } from './types';
+
+export const MENU_LINKS: MenuLink[] = [
+  {
+    title: 'Home',
+    link: '/',
```
jtaaa, line 6: We can make this a `.tsx` in the future and put the icons directly in here if we want to.

client/src/components/Page/index.tsx
```diff
@@ -1,14 +1,15 @@
...
+import Header from 'components/Header';
 
 const styles = (theme: Theme) =>
   createStyles({
     page: {
       display: 'flex',
       flex: 1,
       flexDirection: 'column',
-      alignItems: 'center',
+      alignItems: 'stretch',
```
jtaaa, line 12: This was kinda messing me up because setting `width: 100%` would not include padding unless I used `box-sizing: border-box` but then my vertical padding would also go inside my box.

I think it's more natural to have the children of page `stretch` because `div`s normal fill the width of the container and on mobile this is probably what we want most often. I'm open to changing it back if you don't like this.

client/src/components/Page/index.tsx
```diff
@@ -21,7 +22,18 @@ interface Props extends WithStyles<typeof styles> {
 }
 
 function Page({ classes, children }: Props) {
-  return <div className={classes.page}>{children}</div>;
+  const [isAuthPanelOpen, setIsAuthPanelOpen] = useState(false);
+  const openAuthPanel = () => {
+    setIsAuthPanelOpen(true);
+    console.log({ isAuthPanelOpen });
```
jtaaa, line 28: This is just a stub. I'm thinking the AuthPanel would be mounted on the Page component so it can be opened from any page?

jtaaa:
Oh and this isn't pixel perfect what Josh has in his designs because he uses a lot of strange numbers for padding and spacing (not on 8px grid) so I just kinda used the closest numbers and did what I thought looked good.

jtaaa:
Also when we get to the point of theming, the HorizontalLogo includes text in black which would break a dark theme.

client/src/components/Page/index.tsx
chriskrogh: sgtm!

.gitignore
chriskrogh: yea f that guy

client/src/components/Header/constants.ts
chriskrogh: Interesting thought. I might rather have a `icons.tsx` in the case that this boy gets too big. But we'll cross that road when we get there.

client/src/components/Page/index.tsx
chriskrogh: nit: you might even be able to remove this prop. i think [the default](https://www.w3schools.com/CSSref/css3_pr_align-items.asp) seems to be `stretch`
Up 2 u tho

client/src/components/Page/index.tsx
chriskrogh: Makes sense 2 me

client/src/components/Page/index.tsx
jtaaa: true!

PR #33, opened 2020-07-05
[hefty] Auth Panel UI

chriskrogh:
## What's changed
Added UI for Auth Panel.
Something notable that changed was the maintenance of the Auth Panel is now specific to each page.
i.e. 
Membership page maintains Auth Panel on Membership page
Home page maintains Auth Panel on Home page
Page maintains Auth Panel controlled by the header

The reason for this is because the Membership page can render an Auth Panel with 3 different options whereas every other page (I think) maintains an Auth Panel with just the `login` options.

This means that each "page" that has an Auth Panel will have to render an Auth Panel. 

Feedback on this approach is much appreciated

## UI (if UI has changed)
Mobile
Screen-Recording-2020-07-05-at-1

Desktop
Screen-Recording-2020-07-05-at-5

## Notes
Will create an additional PR for input validation and integration with APIs.
This PR is just for UI.
If there's anything that I missed pls mention and I will address in a PART 2 PR


client/src/pages/Membership/index.tsx
```diff
@@ -1,10 +1,38 @@
...
+
   return (
     <Page>
-      <h1>Membership</h1>
+      <AuthPanel
+        isOpen={isAuthPanelOpen}
+        setIsOpen={setIsAuthPanelOpen}
+        option={authPanelOption}
+      />
+      <Button onClick={() => openAuthPanel(AUTH_PANEL_OPTIONS.LOGIN)}>
```
chriskrogh, line 27: just placeholders for demo purposes. Should be removed when there is actual content

chriskrogh:
> LGTM!
> 
> I'm not a huge fan of having an extra auth panel for the membership page. We'll have 2 mounted on the membership page when we only really need one. Doesn't hurt tho and it simplifies things. + can be refactored later if it's an issue.

Yea I'm really curious what the best way to solve this problem would be... Would it be to also move the header out from `Page` and have it on every individual page?

I tried having 1 auth panel in `Page` but then I had to pass Auth Panel props to `Page` and things started to feel more complicated than they needed to be.

Oh I think I have a better idea. I think Membership is the only page that has to pass other options to the AuthPanel, so we could have a MembershipPage which is different from a Page and have the separate logic in there.

jtaaa:
I was thinking we could use context for it. Wrapping the children of the `Page` in the provider and then any page can access the auth panel context to open it with whatever options it wants. That way it's really shared between all the pages and doesn't have to get re-mounted and re-rendered when we switch pages.

Header could just continue taking the props since it's really high up but deeper components could use context. Prob low priority tho since the current solution works fine.

chriskrogh:
Yea def low priority.
Maybe an AuthPanel context makes sense. Will try em out when I do API integration

PR #34, opened 2020-07-06
refactor

chriskrogh:
## What's changed
Refactored AuthPanel to use context

It feels a lot cleaner to have this mount only once

A lot of the diffs are just boilerplate new context addition

### DIFF
--- a/client/src/context/authPanel/provider.tsx
+++ b/client/src/context/authPanel/provider.tsx
@@ -1,0 +1,36 @@
+import { /*type*/ AuthPanelOption } from 'components/AuthPanel/types';
+
+import React, { useReducer } from 'react';
+import { AuthPanelContext, initialState } from './state';
+import Reducer from './reducer';
+import ActionTypes from './actions';
+
+interface Props {
+  children: React.ReactNode | React.ReactNode[];
+}
+
+const Provider = ({ children }: Props) => {
+  const [state, dispatch] = useReducer(Reducer, initialState);
+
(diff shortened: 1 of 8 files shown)


client/src/context/authPanel/provider.tsx
```diff
@@ -0,0 +1,36 @@
...
+import React, { useReducer } from 'react';
+import { AuthPanelContext, initialState } from './state';
+import Reducer from './reducer';
+import ActionTypes from './actions';
+
+interface Props {
+  children: React.ReactNode | React.ReactNode[];
+}
+
+const Provider = ({ children }: Props) => {
```
jtaaa, line 12: If u type `Provider` as `React.FC` u don't have to worry about defining children in the props

chriskrogh: Whaaa that's sick

chriskrogh:
> Looks really good 💰 I like this 10/10.
> 
> We might wanna make the Page component wrap the routes instead of each route wrapping itself in a Page component.

I'll ticket this for now because we have an under construction page that shouldn't render the header or the auth panel.
But moving the page outside of the routes is a great idea

[trello card](https://trello.com/c/6k1NIy2q)

client/src/context/authPanel/provider.tsx
chriskrogh: That's sick! Will create a refactor PR

PR #35, opened 2020-07-07
Add ShrinkImage component

jtaaa:
Makes image shrink from 1.25 larger size and optionally shift from 100px lower than normal to help with centring.

ShrinkImage

## Notes
This isn't perfect in that it doesn't actually start at the exact center of the screen and maybe doesn't shrink the exact amount from Josh's designs. I think it'll do the trick though.

The size of the image after the animation is decided by the height of the container it is mounted in. It has `max-height` and `max-width` to avoid exceeding the screen limits.


client/src/components/ShrinkImage/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    },
+    'entered-shift': {
+      transform: 'translateY(0) scale(1)',
+    },
+    exiting: {},
+    'exiting-shift': {},
+    exited: {},
+    'exited-shift': {},
+    unmounted: {},
+    'unmounted-shift': {},
```
jtaaa, line 40: These are just stubs so that TS doesn't complain when I index `classes` by `state`.

client/src/components/ShrinkImage/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  classes,
+  shift,
+  ...imgProps
+}: Props) => {
+  return (
+    <Transition in appear timeout={1000}>
+      {(state) => (
+        <img
+          className={classNames(classes.shrinkimage, classes[state], {
+            [classes[`${state}-shift` as ShiftedClassName]]: shift,
```
jtaaa, line 59: Needed this cast because TS widens `${state}-shift` to string :(

chriskrogh: all g in the hood

client/src/components/ShrinkImage/index.tsx
chriskrogh: makes sense

PR #36, opened 2020-07-07
Membership page image

jtaaa:
Added "Are You A Member" image with animation to membership page
AreYouAMember

## Notes
I need to add the animations for the header separately. Any content that comes below the "Are You A Member" image should also have an enter animation with a timeout of 1s so that the image doesn't cover it.


jtaaa:
For now it looks fine up to 1440px desktop but on larger screens, the image is really small. I don't think the solution is to make the image bigger though as it will lose quality.

jtaaa:
Also the image is huge... almost 200kb but we can optimize after if it is a problem

client/src/components/ShrinkImage/index.tsx
```diff
@@ -15,10 +15,13 @@ const styles = (theme: Theme) =>
...
       height: '100%',
-      maxWidth: '80vw',
-      maxHeight: '80vh',
+      maxWidth: '75vw',
+      maxHeight: '75vh',
       margin: `${theme.spacing(4)}px auto`,
       transition: 'transform 300ms ease-in-out',
+      [theme.breakpoints.down('sm')]: {
+        height: 'auto',
+      },
```
jtaaa, line 24: On desktop we can control the size of the image by setting the height of the containing div. We can make it bigger or smaller without hitting the maxWidth and messing up the width/height ratio.

On mobile we let the image scale down appropriately to fit 75% the width of the screen.

chriskrogh: 👍

chriskrogh:
> For now it looks fine up to 1440px desktop but on larger screens, the image is really small. I don't think the solution is to make the image bigger though as it will lose quality.

It looks okay in 4k so this is fine

PR #38, opened 2020-07-12
[UI only] Add SignUp form UI

jtaaa:
Updated the `SignUp` form component to match Josh's designs.
Screen Shot 2020-07-12 at 3 27 04 AM

## Notes
The form doesn't actual do anything yet. This PR contains the UI elements with no state or submission logic.

### DIFF
--- a/client/src/components/forms/SignUp/index.tsx
+++ b/client/src/components/forms/SignUp/index.tsx
@@ -1,47 +1,113 @@
-import React, { useState } from 'react';
-import { signup } from 'utils/data/user';
-import * as M from 'utils/network/errorMessages';
+import React from 'react';
+import {
+  Typography,
+  Theme,
+  createStyles,
+  WithStyles,
+  withStyles,
+  TextField,
+  Box,
+  Divider,
+  Select,
+  MenuItem,
+  FormControl,
+  InputLabel,
+} from '@material-ui/core';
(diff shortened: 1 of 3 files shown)


client/src/components/forms/SignUp/index.tsx
```diff
@@ -1,47 +1,113 @@
...
-      />
-      <div style={{ marginTop: 10 }}>
-        <button onClick={submit}>Create</button>
+      <Typography variant="h4" align="center">
+        Enter Your Information
+      </Typography>
+      <Divider className={classes.titleDivider} />
+      <div className={classes.form}>
+        <Box display="flex">
+          <Box flex={1}>
```
jtaaa, line 46: I like these [`Box`s](https://material-ui.com/components/box/) for small layout styles so I don't have to define a whole class

chriskrogh: Damn that's pretty sick. I think i'll start using them too

client/src/utils/constants.ts
```diff
@@ -2,3 +2,26 @@ export enum ROUTES {
...
+export const SEMESTERS = [
+  '1A',
+  '1B',
+  '2A',
+  '2B',
+  '3A',
+  '3B',
+  '4A',
+  '4B',
+  '5A',
```
chriskrogh, line 24: 5 ers???

PR #39, opened 2020-07-15
Screen Size Breakpoints

chriskrogh:
This was actually merged when I did the construction page, but I think you should be aware of this hook and give feedback on it if you want.

Essentially this hook acts like `breakpoints` in MUI. 
It maintains a piece of state called `ss` (screen size) and this is updated when the breakpoints are crossed.

## What's changed
1. Added `up` and `down` functions.

The reason why we do checks agains the current screen size instead of the width is because it's a lot more expensive to maintain the window's width in state rather than maintaining the breakpoint.

I know so far in our project we've been using one breakpoint (`sm`) but I included all the breakpoints so that we have the option to use them if we need to.

lmk what u think

### DIFF
--- a/client/src/pages/Construction/index.tsx
+++ b/client/src/pages/Construction/index.tsx
@@ -44,7 +44,7 @@ const styles = (theme: Theme) =>
     },
   });
 
-function Home({ classes }: WithStyles<typeof styles>) {
+function Construction({ classes }: WithStyles<typeof styles>) {
   const { down } = useScreenSize();
   const bannerSrc = down(ScreenSize.SM)
     ? 'assets/constr.png'
@@ -80,4 +80,4 @@ function Home({ classes }: WithStyles<typeof styles>) {
   );
 }
 
-export default withStyles(styles)(Home);
+export default withStyles(styles)(Construction);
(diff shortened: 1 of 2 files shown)


client/src/pages/Construction/index.tsx
```diff
@@ -81,4 +80,4 @@ function Home({ classes }: WithStyles<typeof styles>) {
   );
 }
 
-export default withStyles(styles)(Home);
+export default withStyles(styles)(Construction);
```
chriskrogh, line 83: Forgot 2 rename d boy

client/src/utils/hooks/useScreenSize.ts
```diff
@@ -28,7 +29,56 @@ function useScreenSize() {
...
+      case ScreenSize.SM:
+        return ss === ScreenSize.XS || ss === ScreenSize.SM;
+      case ScreenSize.MD:
+        return (
+          ss === ScreenSize.XS || ss === ScreenSize.SM || ss === ScreenSize.MD
+        );
+      case ScreenSize.LG:
+        return (
+          ss === ScreenSize.XS ||
+          ss === ScreenSize.SM ||
```
chriskrogh, line 46: We could do something like `ss !== ScreenSize.XL` here, but I was going for readability here.
Thoughts?

PR #42, opened 2020-07-20
[signup p1] Add Create Account UI

jtaaa:
Added UI element that is shown after a user's account information is inputted to request a password for the account.

Screen Shot 2020-07-19 at 10 45 19 PM

## Notes
Only the UI is added nothing is hooked up yet. Will hook up everything in a separate PR


jtaaa:
I'm not sure why those two empty commits are showing up 🤔

PR #45, opened 2020-07-20
[signup p2] Add sign-up complete UI

jtaaa:
Added the `SignUpComplete` component which we show to the user after they create an account.
Screen Shot 2020-07-19 at 11 04 57 PM

## Notes
Only added UI again 👍  I added some animations but they don't really match Josh's properly (the title and checkmark don't appear, wait, and then slide up, they just slide in). I figure they're better than nothing for now and we can improve them later when we seriously tackle animations.


jtaaa:
I merged all the other parts onto this one and those didn't require re-reviews since their base isn't master but this boy I'll just merge as an admin.

Not sure how this is supposed to be avoided...

PR #46, opened 2020-07-20
[signup p3][⚓] Signup component

jtaaa:
Added the `SignUp` component. Not to be confused with the previous `SignUp` component in the `components/forms` directory which is now renamed to `SignUpForm`.

This is a hefty boy. It's mostly just refactors to make the two form components controlled.

SignUpComponent

## Notes
- Makes `CreateAccountForm` and `SignUpForm` components controlled in that they take their values and setters from props.
- Validation will be done in each component and will only call `onNext` if validation passes.

### DIFF
--- a/client/src/components/SignUp/index.tsx
+++ b/client/src/components/SignUp/index.tsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import SignUpComplete from 'components/SignUpComplete';
 import SignUpForm from 'components/forms/SignUp';
 import CreateAccountForm from 'components/forms/CreateAccount';
+import { signup } from 'utils/data/user';
 
 const SignUp = () => {
   const [firstName, setFirstName] = useState('');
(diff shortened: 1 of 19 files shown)


client/src/components/SignUp/index.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { useState } from 'react';
```
chriskrogh, line 2: this is a super nit (don't take on this comment😂) but we could do 
```
import React, { useState } from 'react';
```
not that big of a deal at all tbh. but maybe for future components?

jtaaa: Yeahh I much prefer it like that. For some reason TypeScript will auto import it into a separate import sometimes. I usually switch em out but I missed this guy.

Will fix and use my admin privileges to merge without re-review 💪

jtaaa:
🚫 `/* type */` 🚫🚫!

PR #47, opened 2020-07-20
[signup p4] Create account validation

jtaaa:
Added some validation to the `CreateAccountForm`. It will now only call `onNext` if its validation passes. Validation includes:

1. Password is valid (at least 8 characters)
2. Password and re-entered password matches

CreateAccountValidation

## Notes
I had to copy and modify the `isPassword` util from the server-side since we don't have a great way to share code between them.

### DIFF
--- a/client/src/components/forms/CreateAccount/constants.ts
+++ b/client/src/components/forms/CreateAccount/constants.ts
@@ -1,0 +1,9 @@
+export const PASSWORDS_DONT_MATCH_ERROR =
+  'The passwords you entered do not match.';
+
+export const INVALID_PASSWORD_ERROR =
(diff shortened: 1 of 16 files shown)


client/src/components/forms/CreateAccount/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const PASSWORDS_DONT_MATCH_ERROR =
+  'The passwords you entered do not match.';
+
+export const INVALID_PASSWORD_ERROR =
+  'Your password must be at least 8 characters long.';
+
+export type CreateAccountError =
```
chriskrogh, line 7: Big fan of these

PR #48, opened 2020-07-20
[signup p5] Add SignUpForm validation

jtaaa:
Added validation to `SignUpForm`.

SignUpFormValidation

## Notes
We can tweak the validation for first name and last name. Right now it's `length > 0` and only contains alphabetic characters, spaces and dashes.

### DIFF
--- a/client/src/components/forms/SignUp/index.tsx
+++ b/client/src/components/forms/SignUp/index.tsx
@@ -74,42 +74,15 @@ function SignUpForm({
   setFaculty,
   onNext,
 }: Props) {
-  const [showFirstNameError, setShowFirstNameError] = useState(false);
-  const [showLastNameError, setShowLastNameError] = useState(false);
-  const [showStudentNumberError, setShowStudentNumberError] = useState(false);
-  const [showEmailError, setShowEmailError] = useState(false);
-  const [showSemesterError, setShowSemesterError] = useState(false);
-  const [showFacultyError, setShowFacultyError] = useState(false);
   const [triedToSubmit, setTriedToSubmit] = useState(false);
 
-  useEffect(() => {
-    if (triedToSubmit && !isName(firstName)) setShowFirstNameError(true);
-    else setShowFirstNameError(false);
-
-    if (triedToSubmit && !isName(lastName)) setShowLastNameError(true);
-    else setShowLastNameError(false);
-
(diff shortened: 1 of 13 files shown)


client/src/components/forms/SignUp/index.tsx
```diff
@@ -64,6 +74,57 @@ function SignUpForm({
   setFaculty,
   onNext,
 }: Props) {
+  const [showFirstNameError, setShowFirstNameError] = useState(false);
+  const [showLastNameError, setShowLastNameError] = useState(false);
+  const [showStudentNumberError, setShowStudentNumberError] = useState(false);
+  const [showEmailError, setShowEmailError] = useState(false);
```
chriskrogh, line 80: Is there a reason why we want to maintain these values in state instead of computing them?
E.g.
If we already maintain `firstName` in state, and we can compute whether or not we want to show the error, is there a reason why we would store the error state?

client/src/components/forms/SignUp/index.tsx
```diff
@@ -79,6 +140,8 @@ function SignUpForm({
               fullWidth
               value={firstName}
               onChange={(e) => setFirstName(e.target.value)}
+              error={showFirstNameError}
+              helperText={showFirstNameError ? FIRST_NAME_ERROR : undefined}
```
chriskrogh, line 144: (Similar to comment above)
I think all of these could be computed.
E.g.
```
helperText={showFirstNameError(firstName) && FIRST_NAME_ERROR}
```

client/src/components/forms/SignUp/index.tsx
jtaaa: Nope there's no reason 😨 Switched it over to computing them in the render

client/src/components/forms/SignUp/index.tsx
jtaaa: 💯 %

client/src/components/forms/SignUp/index.tsx
```diff
@@ -64,6 +74,30 @@ function SignUpForm({
   setFaculty,
   onNext,
 }: Props) {
+  const [triedToSubmit, setTriedToSubmit] = useState(false);
+
+  const showFirstNameError = triedToSubmit && !isName(firstName);
```
chriskrogh, line 79: Ooh boy 😍
str8 🔥

PR #50, opened 2020-07-20
[signup p7] Signup request

jtaaa:
The `SignUp` component will now actually sign up the user from the `CreateAccountForm` by sending a request with the info to the backend.

The component is basically complete now except for a failed signup state. It just needs to be dropped into a `JoinACS` or other name component and displayed after the signup button on the Membership page is pressed¡

### DIFF
--- a/client/src/components/SignUp/index.tsx
+++ b/client/src/components/SignUp/index.tsx
@@ -3,6 +3,7 @@ import { useState } from 'react';
 import SignUpComplete from 'components/SignUpComplete';
 import SignUpForm from 'components/forms/SignUp';
 import CreateAccountForm from 'components/forms/CreateAccount';
+import { signup } from 'utils/data/user';
 
 const SignUp = () => {
   const [firstName, setFirstName] = useState('');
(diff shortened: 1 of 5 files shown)


client/src/components/SignUp/index.tsx
```diff
@@ -18,6 +19,21 @@ const SignUp = () => {
   const [isInfoEntered, setIsInfoEntered] = useState(false);
   const [isSignupComplete, setIsSignupComplete] = useState(false);
 
+  const onSignUp = async () => {
+    const watIAMUserId = email.split('@')[0];
+    await signup({
+      firstName,
+      lastName,
+      studentNumber: +studentNumber,
```
chriskrogh, line 27: What does this mean? Looks cool

client/src/context/user/state.ts
```diff
@@ -1,14 +1,23 @@
 import { createContext } from 'react';
-import { User } from 'types/user';
+import { User, MEMBERSHIP_STATUS } from 'types/user';
+
+const userState: User = {
```
chriskrogh, line 4: Good refactor

client/src/types/user.ts
```diff
@@ -16,9 +16,11 @@ export type UserData = Omit<Credentials, 'password'> & {
   lastName: string;
   watIAMUserId: string;
   studentNumber: number;
+  semester: string;
+  faculty: string;
   picture?: string;
   membershipStatus: MEMBERSHIP_STATUS;
-  isAdmin: string;
+  isAdmin: boolean;
```
chriskrogh, line 23: Yessir

client/src/components/SignUp/index.tsx
jtaaa: hahaha it's to convert `studentNumber` from `string` to `number`.

Also `studentNumber` is must be a valid number string since we use `type="number"` on its `TextField`. Otherwise we could run into studentNumber being converted to `NaN` here if it isn't a valid number.

PR #51, opened 2020-07-20
[tiny] Fix lodash dependabot alert

jtaaa:
Added latest `lodash` dependency to client and server. Hopefully this fixes the dependable alert, not 100% sure it will tho 😁


jtaaa:
Looks like u got the dependabot thing to make a PR so I'll close this

PR #54, opened 2020-07-20
Membership Options 3

chriskrogh:
## What's changed
Refined UI
1. Fixed width for an option
2. Shows description on mobile
3. Transition from bottom to top

## UI (if UI has changed)
desktop
Screen-Recording-2020-07-20-at-7

mobile
Screen Shot 2020-07-20 at 7 24 32 PM

### Todo
Do transition for arrow

### DIFF
--- a/client/src/components/lists/MembershipOption/option.tsx
+++ b/client/src/components/lists/MembershipOption/option.tsx
@@ -16,6 +16,7 @@ const styles = (theme: Theme) =>
       flexDirection: 'column',
       alignItems: 'center',
       padding: theme.spacing(4),
+      cursor: 'pointer',
       [theme.breakpoints.up('sm')]: {
         width: '10%',
       },
@@ -39,12 +40,12 @@ interface Props extends WithStyles<typeof styles> {
 
 function Option({ classes, theme, option, isLast }: Props) {
   const [showOptions, setShowOptions] = useState(false);
-  const [showDescription, setShowDescription] = useState(false);
+  const [isHovering, setIsHovering] = useState(false);
   const { setOption, setIsOpen } = useContext(AuthPanelContext);
 
   useEffect(() => {
     setTimeout(() => setShowOptions(true), 1000);
-  });
+  }, []);
 
   const openAuthPanel = () => {
     if (option.authPanelOption) {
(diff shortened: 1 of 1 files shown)


client/src/components/lists/MembershipOption/option.tsx
```diff
@@ -28,14 +32,20 @@ const styles = (theme: Theme) =>
...
 
-function Option({ classes, option, isLast }: Props) {
+function Option({ classes, theme, option, isLast }: Props) {
+  const [showOptions, setShowOptions] = useState(false);
   const [showDescription, setShowDescription] = useState(false);
   const { setOption, setIsOpen } = useContext(AuthPanelContext);
 
+  useEffect(() => {
+    setTimeout(() => setShowOptions(true), 1000);
+  });
```
jtaaa, line 47: I think this boy will run on every render since u didn't pass a dependency array

```suggestion
  useEffect(() => {
    setTimeout(() => setShowOptions(true), 1000);
  }, []);
```

An empty deps array will make it run only on first render 💯

client/src/components/lists/MembershipOption/option.tsx
```diff
@@ -44,28 +54,30 @@ function Option({ classes, option, isLast }: Props) {
...
+          variant="h4"
+          align="center"
+          className={classnames(classes.text, classes.title)}
+        >
+          {option.title}
         </Typography>
-      )}
-      <BWButton onClick={openAuthPanel}>{option.cta}</BWButton>
-    </div>
+        {(useMediaQuery(theme.breakpoints.down('sm')) || showDescription) && (
```
jtaaa, line 72: I've never had a good reason to use hooks inside the JSX but I kinda like this 👍 👍

webveloperchris: Yea. If we start using the same breakpoint in the same file, then we should extract to a variable.
But this is only being used here I think...

chriskrogh:
> Looks good to me except the useEffect deps 💯
> 
> What do u think about making the whole thing clickable and once u hover, adding the hover to the `BWButton` programmatically
> 
> ```
> // for cases where we want to add hover style programmatically
> export const HoverClassApplied = () => (
>   <BWButton classes={{ root: 'hover' }}>BWButton with Hover Styles</BWButton>
> );
> ```

Lol I keep forgetting to switch to my personal GH account.

I'm down to try em!

chriskrogh:
## New UI
Whole option is clickable. Hover state sets when user focusses on option
Screen-Recording-2020-07-22-at-4

PR #55, opened 2020-07-23
[easy] Member Benefits

chriskrogh:
This PR should be super light to review. Nothing too fancy / crazy.

## What's changed
Added UI fro Member Benefits

## UI (if UI has changed)
Screen Shot 2020-07-22 at 9 33 39 PM

I haven't put this together with the sign up form, but maybe we should take another look at the `integration` tasks since we're done individual sections?

Lmk how u wanna put this all together. I can help out if you want😅

### DIFF
--- a/client/src/components/Benefits/benefits.ts
+++ b/client/src/components/Benefits/benefits.ts
@@ -1,0 +1,18 @@
+export default [
+  {
+    iconColor: 'rgb(0, 0, 145)',
+    text: 'One ACS t-shirt per year',
+  },
+  {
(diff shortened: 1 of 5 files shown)


client/src/components/Benefits/benefits.ts
```diff
@@ -0,0 +1,18 @@
+export default [
```
jtaaa, line 1: if you type this fella as `Benefit[]` you won't have to elsewhere

chriskrogh: Good 2 know! Will update

PR #58, opened 2020-07-28
Add SignUpPage on /membership/signup route

jtaaa:
Added da `SignUpPage`

Screen Shot 2020-07-27 at 8 23 15 PM
localhost_3000_membership_signup

## Notes
I had to add a key to the Fade component to get it to work properly. I think React was confusing it for the other `Fade` component from `/howtojoin` and not unmounting in remounting it on navigation. Not sure but the key prop fix works.

### DIFF
--- a/client/src/components/forms/SignUp/index.tsx
+++ b/client/src/components/forms/SignUp/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import {
   Typography,
   Theme,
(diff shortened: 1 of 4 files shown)


jtaaa:
Done 🧇

client/src/components/forms/SignUp/index.tsx
```diff
@@ -100,7 +103,7 @@ function SignUpForm({
 
   return (
     <div>
-      <Typography variant="h4" align="center">
+      <Typography variant="h4" align="center" className={classes.title}>
```
chriskrogh, line 106: 💯

jtaaa: thinking about it now i could have just used one of these boys `<em>...</em>`

PR #62, opened 2020-07-31
Verify Info PART 2

chriskrogh:
## What's changed
Added the component that has verify your information.

Right now the default values on the form are set from the user in context. 
The user can modify these values using the form.
When they click submit, the validation from utils is run (`isName`, `isUWEmail`...)
If everything is good, it sets these updated values to the user in context.

## UI (if UI has changed)
from storybook
Screen Shot 2020-07-30 at 10 34 56 PM

## Notes
Next I'll be creating this
Screen Shot 2020-07-30 at 7 53 48 PM

And then I'll put things together🤙

### DIFF
--- a/client/src/components/VerifyInfo/index.tsx
+++ b/client/src/components/VerifyInfo/index.tsx
@@ -1,0 +1,102 @@
+import type { WithStyles, Theme } from '@material-ui/core/styles';
+
+import React, { useState, useContext } from 'react';
+import { withStyles, createStyles } from '@material-ui/core/styles';
+import { Fade, Box, Typography } from '@material-ui/core';
+import { UserContext } from 'context/user/state';
+import UserInfoForm from 'components/forms/UserInfo';
+import BWButton from 'components/buttons/BWButton';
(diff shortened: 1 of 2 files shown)


client/src/components/VerifyInfo/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  return (
+    <div>
+      <Fade appear in timeout={800}>
+        <Box display="flex" flexDirection="column" alignItems="center">
+          <Typography variant="h3" align="center">
+            Verify Your Information
+          </Typography>
+          <div className={classes.subtitle}>
```
jtaaa, line 64: just interested

I really like boxes here so that you can see the layout of the JSX without having to scroll up into the classes.
There's also the argument that it makes things more cluttered and splits styles into two places.

what do u think?

chriskrogh: Yea it's just been a habit of mine to use divs. Boxes look cool. 
Will try to use them from now on.
But we can also add a refactor to swap all possible divs for boxes

PR #64, opened 2020-08-03
[ez] Fix non-master branch alert in deploy script

jtaaa:
## What's changed
The deploy script no longer compares `"branch"` to `"master"` to determine if we're on the master branch ! 😀 

## UI (if UI has changed)
No change

## Notes
Use the variable value itself and not the string literal for comparisons 👍


jtaaa:
btw I already deployed this branch to make sure it works 🥯

PR #65, opened 2020-08-03
[fix][ez] Add a wrapper around img in ShrinkImage

jtaaa:
## What's changed
The `ShrinkImage` works in production builds  🎉 

## UI (if UI has changed)
`prodUI === localUI`

## Notes
Super weird bug that took way to long to figure out 😅 Not sure exactly why wrapping the return of the child render function works or why I can't find anything about it online but maybe we just wrap all our returns like that in single `div`s.

Related to [this trello item](https://trello.com/c/5je62pet/52-shrinkimage-doesnt-show-up-in-production)


jtaaa:
btw I already deployed this branch to make sure it works 🥯

PR #67, opened 2020-08-03
[ez] Add ShrinkImage to and use HomePage

jtaaa:
## What's changed
The construction page is replaced with the home page which is now no longer empty.
It now contains the `ShrinkImage` with our ACS Flags logo image thing.

## UI
Screen Shot 2020-08-03 at 5 55 07 PM


jtaaa:
> We can merge this if we change the `Home` route to `/home` and change all references to `/` to `/home` for now.
> 
> When we're ready to launch we can change all occurrences of `/home` to `/`. What do you think?

sounds good to me

jtaaa:
Done. Construction page is on `/` and the header is now using the `ROUTES` enum so it links to the `Home` page properly ¡

PR #70, opened 2020-08-05
Renew Membership Complete

chriskrogh:
## What's changed
1. Renamed `SignUpComplete` -> `Complete` because we're using the same component for renew membership.
2. Added a variant prop. The only difference between the components is the title suffix.
I was thinking to extract the title, description and button text to have the ability to show that component in different scenarios (like when we implement voting) but we can make that refactor when we get there.
It might also be the case that we just add another variant like `voting` so that the new `titleSuffix` will be voting and everything else is the same. Don't wanna think too far ahead but those are some thoughts.
3. Linked the button to the events page. Even though it's not built yet, we can just leave that there 4 now. I think by the time ppl start using the site we'd have something on that page since that's next on the chopping block.

## UI (if UI has changed)
Screen Shot 2020-08-05 at 9 28 20 AM

## Notes
Gonna link things up


client/src/components/Complete/index.tsx
```diff
@@ -26,11 +33,13 @@ const SignUpComplete = () => {
       </Fade>
       <Fade appear in timeout={1000}>
         <Box marginTop={6} display="flex" justifyContent="center">
-          <BWButton>View Upcoming Events</BWButton>
+          <BWButton component={Link} to="/events">
```
jtaaa, line 36: We can use the `ROUTES` enum here. I just swapped the Header Menu links to use them 🚀

chriskrogh: Updated!

PR #71, opened 2020-08-05
Renew Membership Flow

chriskrogh:
## What's changed
Completed the renew membership flow

## UI (if UI has changed)
Desktop
Screen-Recording-2020-08-05-at-1

Mobile
Screen-Recording-2020-08-05-at-1 (1)

Looks good on tablet too😂 I checked

## Notes
I'll work on backend integration sometime. 
But we should work on edge cases.
E.g. if a user tries to sign up but their email / student id is already in use.
same thing with membership renewal

Need to start handling those from UI side

### DIFF
--- a/client/src/components/VerifyInfo/index.stories.tsx
+++ b/client/src/components/VerifyInfo/index.stories.tsx
@@ -8,10 +8,10 @@ export default {
 
 const baseProps = {};
 
-export const Base = () => <VerifyInfo {...baseProps} />;
+export const Base = () => <VerifyInfo {...baseProps} onVerify={() => {}} />;
 
 export const LimittedWidth = () => (
   <div style={{ width: '600px' }}>
-    <VerifyInfo {...baseProps} />
+    <VerifyInfo {...baseProps} onVerify={() => {}} />
   </div>
 );
--- a/client/src/components/VerifyInfo/index.stories.tsx
+++ b/client/src/components/VerifyInfo/index.stories.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import VerifyInfo from '.';
+import { action } from '@storybook/addon-actions';
 
 export default {
   component: VerifyInfo,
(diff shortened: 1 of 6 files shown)


client/src/components/VerifyInfo/index.stories.tsx
```diff
@@ -8,10 +8,10 @@ export default {
 
 const baseProps = {};
 
-export const Base = () => <VerifyInfo {...baseProps} />;
+export const Base = () => <VerifyInfo {...baseProps} onVerify={() => {}} />;
```
jtaaa, line 11: We can use `action('onVerify')` for the `onVerify` prop of these stories, imported from `@storybook/addon-actions` I think.

That way we'll get a little action notification in the addons panel 🚀

client/src/pages/Membership/VerifyInfo/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+const styles = (theme: Theme) =>
+  createStyles({
+    container: {
+      display: 'flex',
+      flexDirection: 'column',
+      alignItems: 'center',
+      width: `calc(100% - ${2 * theme.spacing(3)}px)`,
+      padding: `0 ${theme.spacing(3)}px`,
+    },
+    contentContainer: {},
```
jtaaa, line 17: this fellow isn't needed?

client/src/components/VerifyInfo/index.tsx
```diff
@@ -19,7 +19,11 @@ const styles = (theme: Theme) =>
     },
   });
 
-function VerifyInfo({ classes }: WithStyles<typeof styles>) {
+type Props = WithStyles<typeof styles> & {
+  onVerify: () => void;
```
jtaaa, line 23: Will `onVerify` need to be async to query the backend?

client/src/pages/Membership/VerifyInfo/index.tsx
chriskrogh: No he is not! Thanks for the catch!

client/src/components/VerifyInfo/index.tsx
chriskrogh: Yessir. Will update in next PR

PR #72, opened 2020-08-10
[⚓] Handle duplicate key errors for SignUp form

jtaaa:
## What's changed  
If the signup request throws an error because of attempting to add a user with a duplicated unique database field we now handle it gracefully on the frontend by displaying an error message saying "This value is already linked to an existing user" 🚀 

## UI (if UI has changed)
Screen Shot 2020-08-10 at 3 43 39 PM
Screen Shot 2020-08-10 at 3 43 31 PM

## Notes
I expanded the error functionality of the SignUp form on a whole so that if there are other errors other than regular input errors we can include those as well.
Also added the `RequestError` class which stores the data and original errors along with an error message.


client/src/components/SignUp/index.tsx
```diff
@@ -15,22 +18,34 @@ const SignUp = () => {
...
+      });
+      setIsSignupComplete(true);
+    } catch (err) {
+      const error = err as RequestError<{
+        duplicateKeys?: Partial<FieldValues>;
+      }>;
+      if (error.data.duplicateKeys) {
+        setIsInfoEntered(false);
+        setDuplicateKeys(error.data.duplicateKeys);
+      }
```
jtaaa, line 47: Right now I'm not rethrowing the error even if it does not contain the `duplicateKeys` field (maybe error'd for a different reason). So the request will fail but the client won't crash. I guess we really should throw if it's not an error we explicitly handle but we also need to set up Error boundaries so the whole app doesn't crash.

chriskrogh: Good thinking. This looks good
I think we could have an error boundary added in the near future. 
Will ticket

client/src/components/SignUp/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  const errors: Partial<FieldValues> = {};
+  (Object.keys(duplicatedKeys) as Array<FieldKeys>).forEach((key) => {
+    const dupe = duplicatedKeys[key]?.toString();
+    const value = values[key]?.toString();
+    if (dupe && value && dupe === value) {
+      errors[key] = 'This value is already linked to an existing user';
+    }
+  });
+  return errors;
+};
```
chriskrogh, line 16: This is pretty sick!

client/src/components/forms/UserInfo/constants.ts
```diff
@@ -1,9 +1,11 @@
-export const STUDENT_NUMBER_ERROR = 'Student number must be 8 digits long.';
-export const FIRST_NAME_ERROR =
-  'First name is required and must not contain numbers or special characters.';
-export const LAST_NAME_ERROR =
-  'Last name is required and must not contain numbers or special characters.';
-export const EMAIL_ERROR =
-  'Email must be your uwaterloo.ca email address provided by the school.';
-export const SEMESTER_ERROR = 'Semester is required.';
-export const FACULTY_ERROR = 'Faculty is required.';
+export const INPUT_ERRORS = {
```
chriskrogh, line 1: Big fan of this construction

client/src/components/forms/UserInfo/index.tsx
```diff
@@ -48,15 +50,16 @@ function UserInfoForm({
   setSemester,
   faculty,
   setFaculty,
+  errors,
   triedToSubmit,
 }: Props) {
-  const showFirstNameError = triedToSubmit && !isName(firstName);
-  const showLastNameError = triedToSubmit && !isName(lastName);
-  const showStudentNumberError =
+  const showFirstNameInputError = triedToSubmit && !isName(firstName);
```
chriskrogh, line 56: Good renaming

PR #73, opened 2020-08-25
VotingModel and update voting stage route

jtaaa:
## What's changed
We now have a VotingModel with the static `getVotingDoc` method which should always be used instead of `.find()`, `.findOne()`, etc since it only contains one document with global voting data (right now only the voting `stage`).

I also added the PATCH `/api/voting/stage` route for admins to update the stage.

The auth middleware function now takes `AuthOptions` and returns the middleware with those options set. This is useful for customizing the middleware. Now we can do `auth({ isAdmin: true })` to require that the user is authenticated and an admin.
I refactored all the code using `auth` as middleware to `auth()`. 

## Notes
- Not a huge fan of this method of creating a whole table to store one doc but it works!
- Not sure if you're planning on nesting the `/nominations` routes in voting but we could

### DIFF
--- a/src/models/voting.ts
+++ b/src/models/voting.ts
@@ -10,7 +10,7 @@ const VotingSchema = new mongoose.Schema({
   },
 });
 
-VotingSchema.statics.getVotingDoc = async () => {
+VotingSchema.statics.getDoc = async () => {
   let votingDoc = await VotingModel.findOne({});
   if (!votingDoc) {
     votingDoc = await VotingModel.create({ stage: VOTING_STAGE.Closed });
@@ -22,7 +22,7 @@ interface VotingDoc extends mongoose.Document {
   stage: VOTING_STAGE;
 }
 interface Voting extends mongoose.Model<VotingDoc> {
-  getVotingDoc: () => Promise<VotingDoc>;
+  getDoc: () => Promise<VotingDoc>;
 }
 
 const VotingModel = mongoose.model<VotingDoc, Voting>('Voting', VotingSchema);
--- a/src/models/voting.ts
+++ b/src/models/voting.ts
@@ -1,0 +1,30 @@
+import mongoose from 'mongoose';
+import { VOTING_STAGE } from '../types/voting';
+import { getStringEnumValues } from '../utils/enum';
+
+const VotingSchema = new mongoose.Schema({
+  stage: {
+    type: String,
+    enum: getStringEnumValues(VOTING_STAGE),
+    required: true,
+  },
+});
+
+VotingSchema.statics.getVotingDoc = async () => {
+  let votingDoc = await VotingModel.findOne({});
+  if (!votingDoc) {
+    votingDoc = await VotingModel.create({ stage: VOTING_STAGE.Closed });
+  }
+  return votingDoc;
+};
+
+interface VotingDoc extends mongoose.Document {
+  stage: VOTING_STAGE;
+}
+interface Voting extends mongoose.Model<VotingDoc> {
+  getVotingDoc: () => Promise<VotingDoc>;
+}
+
+const VotingModel = mongoose.model<VotingDoc, Voting>('Voting', VotingSchema);
+
+export default VotingModel;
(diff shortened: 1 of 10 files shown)


src/models/voting.ts
```diff
@@ -0,0 +1,30 @@
+import mongoose from 'mongoose';
+import { VOTING_STAGE } from '../types/voting';
+import { getStringEnumValues } from '../utils/enum';
+
+const VotingSchema = new mongoose.Schema({
+  stage: {
+    type: String,
+    enum: getStringEnumValues(VOTING_STAGE),
```
webveloperchris, line 8: this is cool

src/models/voting.ts
```diff
@@ -0,0 +1,30 @@
...
+  },
+});
+
+VotingSchema.statics.getVotingDoc = async () => {
+  let votingDoc = await VotingModel.findOne({});
+  if (!votingDoc) {
+    votingDoc = await VotingModel.create({ stage: VOTING_STAGE.Closed });
+  }
+  return votingDoc;
+};
```
webveloperchris, line 19: I'm not too huge of a fan of this. It seems to me like `!votingDoc` will only be true once and false every other time. This code makes me fee like for some reason `votingDoc` will not be available sometimes and we'll have to make a new one. Since we (the devs) are the only ones who have to deal with the case where the voting doc doesn't exist, maybe we can write a dev route that creates the voting doc if it doesn't already exist for the devs to use. Rather than have that additional check that is always gonna be false on each call.

What do you think?

src/routers/user/index.ts
```diff
@@ -75,7 +75,7 @@ router.post('/thirdPartyAuth', async (req: Request, res: Response) => {
 // logout
 router.post(
   '/logout',
-  auth,
+  auth(),
```
webveloperchris, line 78: Just to clarify, does invoking this with no param mean that I am not admin? how would the admin routes be triggered?

src/utils/enum.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export const getStringEnumKeys = (E: any) =>
+  Object.keys(E).filter((k) => typeof E[k as any] === 'string');
+
+export const getStringEnumValues = (E: any) =>
+  getStringEnumKeys(E).map((k) => E[k as any]);
```
webveloperchris, line 6: Big fan. Pretty cool

src/routers/user/index.ts
jtaaa: Invoking with not params is exactly as it was before so no admin checking.
Invoking with `{ isAdmin: true }` requires `isAdmin` set to true on the user doc to use route.

src/models/voting.ts
jtaaa: I don't think the additional check is a big problem. We are checking if the doc exists each time even though only it will only not exist on the first run on a fresh DB but in terms of performance, it shouldn't be any different at all. And this way we don't have to run an extra command or hit a dev route before we can get up and running with the app.

If it's about clarity how about if I changed the name to `getOrCreateVotingDoc`?

src/utils/enum.ts
jtaaa: 🚀 
I find it's a bit strange that enums don't have a built-in method for this. Kinda makes me lean towards wanting to use object literals and const assertions.

src/models/voting.ts
chriskrogh: It's actually fine😅 Just feels a lil weird but it's probably okay
If anything you can probably rename to `getDoc` because they'll only ever be 1 and we have to call it on the voting model anyway.

jtaaa: Sounds good to me! We can even get rid of the Model completely and just export the single doc but I'll just update the name for now

jtaaa: Sounds good to me! We can even get rid of the Model completely and just export the single doc but I'll just update the name for now

PR #74, opened 2020-08-27
Login flow

chriskrogh:
## What's changed
Integrated the log in flow with the back end. Also, because you can log in with both your email and your WatIAM, I renamed that variable from `email` to `id`.

## UI (if UI has changed)
Screen-Recording-2020-08-27-at-5

## Notes
Will be doing `renew membership` and `membership status` next


src/models/user.ts
```diff
@@ -105,11 +105,10 @@ UserSchema.statics.validatePassword = async (
...
 
-UserSchema.statics.findByCredentials = async (
-  email: string,
-  password: string,
-) => {
-  const user = await UserModel.findOne({ email });
+UserSchema.statics.findByCredentials = async (id: string, password: string) => {
+  const userWithEmail = await UserModel.findOne({ email: id });
+  const userWithWatIAM = await UserModel.findOne({ watIAMUserId: id });
+  const user = userWithEmail ? userWithEmail : userWithWatIAM;
```
jtaaa, line 111: We can merge these two DB requests with the `$or` operator.

Also, this would've been a good place to use the `??` operator but it won't be necessary if we merge the DB requests

client/src/types/user.ts
```diff
@@ -7,11 +7,12 @@ export enum MEMBERSHIP_STATUS {
...
 
 export type Credentials = {
-  email: string;
+  id: string;
   password: string;
 };
 
-export type UserData = Omit<Credentials, 'password'> & {
+export type UserData = {
+  email: string;
```
jtaaa, line 15: Make sense to get `Credentials` out of `UserData`! 💯

src/models/user.ts
chriskrogh: I kind of bricked this PR by merging the other 2 in prematurely. However I did update that code here: 782323bff3f0fdf32eefaa498192cc945432b3b1

chriskrogh:
@jtaaa I'm not sure if you wanna review again because I accidentally merged the other 2 PR's into this one. 
But I made the DB `or` change. lmk if this is g2g

PR #75, opened 2020-08-27
Authenticated Renders

chriskrogh:
## What's changed
Not sure if you noticed in https://github.com/uwaterlooacs/acs/pull/74 but when the user authenticates, we don't want them to see log in / sign up buttons. This PR hides those buttons from the header and from the membership page.

## UI (if UI has changed)
Refer to https://github.com/uwaterlooacs/acs/pull/74

### DIFF
--- a/client/src/components/Header/Menu.tsx
+++ b/client/src/components/Header/Menu.tsx
@@ -33,7 +33,6 @@ interface Props extends WithStyles<typeof styles> {
 
 function Menu({ classes, currentPathname, onLoginClicked }: Props) {
   const { user } = useContext(UserContext);
-  const authenticated = user._id !== '0';
 
   return (
     <div className={classes.container}>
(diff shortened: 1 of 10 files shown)


client/src/components/Header/Menu.tsx
```diff
@@ -31,6 +32,9 @@ interface Props extends WithStyles<typeof styles> {
 }
 
 function Menu({ classes, currentPathname, onLoginClicked }: Props) {
+  const { user } = useContext(UserContext);
+  const authenticated = user._id !== '0';
```
jtaaa, line 36: Hmmm maybe we should make `user` possibly undefined or null in the `UserContext` and null/undefined in the initial state. Is there any reason it's populated when the user isn't signed in?

chriskrogh: I'm 100% down for that actually. That would make our lives a lot easier. Will implement

jtaaa:
If you wanna do it in a separate PR I can approve this boy! To avoid conflicts and stuff with the other branches

client/src/components/Header/Menu.tsx
chriskrogh: updated in 48b9d10776c25b5fc24a8411c69180091fed98bf

PR #77, opened 2020-08-28
Nomination and Position models and relationships

chriskrogh:
## What's changed
Created models for `Nomination` and `Position` based on the [doc](https://docs.google.com/document/d/1DYOEQTWimj2cl5EO8Tmq7rpl_dRdEEcayrojUdZ94d8/edit?usp=sharing)

I also defined the relationships between the 3 models:
1. A Position **has an** occupant (User)
2. A Nomination **has a** position (Position)
A Nomination **has a** candidate (User)

## Notes
It's important to know that when we define relationships between models, we have to specify the relationship on both models. The one who owns and the one who is owned.


src/models/user.ts
```diff
@@ -72,10 +73,16 @@ const UserSchema = new mongoose.Schema(
   },
 );
 
-UserSchema.virtual('event', {
-  ref: 'Event',
```
chriskrogh, line 76: This was from the `mern-auth` template. Can be removed no prob

src/models/position.ts
```diff
@@ -0,0 +1,44 @@
...
+    required: true,
+    ref: ModelRefs.USER,
+  },
+});
+
+PositionSchema.virtual('nomination', {
+  ref: ModelRefs.NOMINATION,
+  localField: '_id',
+  foreignField: 'position',
+});
```
jtaaa, line 40: Cool so this will give us an array of related nominations on a `PositionDoc` without us having to add them to the Model 10/10

src/models/types.ts
```diff
@@ -0,0 +1,5 @@
+export enum ModelRefs {
+  USER = 'User',
+  POSITION = 'Position',
+  NOMINATION = 'Nomination',
+}
```
jtaaa, line 5: Cool!

src/models/user.ts
```diff
@@ -72,10 +73,16 @@ const UserSchema = new mongoose.Schema(
...
+UserSchema.virtual('position', {
+  ref: ModelRefs.POSITION,
   localField: '_id',
-  foreignField: 'owner',
+  foreignField: 'occupant',
```
jtaaa, line 79: Should this be `justOne: true`?

src/models/position.ts
jtaaa: should we call this `nominations` plural?

src/models/user.ts
```diff
@@ -72,10 +73,16 @@ const UserSchema = new mongoose.Schema(
...
+});
+
+UserSchema.virtual('nomination', {
+  ref: ModelRefs.NOMINATION,
+  localField: '_id',
+  foreignField: 'candidate',
```
jtaaa, line 85: Maybe this should be plural too?

src/models/user.ts
jtaaa: Big up @`mern-auth`

src/models/position.ts
chriskrogh: Yea this is a bit confusing and I always mix up what this means. (I'll add a comment)
Essentially it means that a `Nomination` has a property called `position` that references this `Position` model.

src/models/user.ts
chriskrogh: I think the schema of `Position` suggests that there will only ever be one occupant. 
What does `justOne: true` do?

src/models/user.ts
chriskrogh: I don't think so. Every nomination can have 1 candidate. If you want to nominate more people for the same position you'd have to create new nominations.

src/models/position.ts
jtaaa: Hmm isn't what you described there the position ref on the Nominations schema? We can have a `Nomination` have a property called `position` that references the `Position` model without this code.

Wouldn't this code actually add a virtual `nomination` array property to the `Position` docs that contains many references to Nominations?

chriskrogh: Honestly I should do some reading into what this code does because I'm not too sure. I thought it was necessary in defining any relationship😂
But now I'm second guessing

src/models/user.ts
jtaaa: It'll make the virtual `position` property on User docs a single doc or null. As documented [here](https://mongoosejs.com/docs/api/schema.html#schema_Schema-virtual).

src/models/position.ts
chriskrogh: Bro I think you're right. we 100% don't need this code🤣

src/models/user.ts
jtaaa: I'm pretty sure here you are defining a virtual `nomination` property on `User` docs that is an array of references to `Nomination` docs. It will include all `Nomination` docs with `candidate` field matching the `User` doc's `_id`.

src/models/position.ts
jtaaa: 🤣  I think it could be useful to have the virtual properties! but yeah not strictly necessary

src/models/user.ts
chriskrogh: Yeaa this makes sense. updated in 775aec3f4620d07c84cc9b51685b62bcfd93f732

src/models/user.ts
chriskrogh: updated in 775aec3f4620d07c84cc9b51685b62bcfd93f732.

chriskrogh:
I think it would be useful to keep track of nominations on a user so that a user can easily see all of the positions they've been nominated for.

PR #78, opened 2020-08-28
[misleadingly big] Create Nominations

chriskrogh:
## What's changed
In this PR, we add the route to create nominations.

However, to create a nomination, you need positions to exist first. So I created a `devRoute` for populating the db with positions (if they don't already exist). We can consider this like a startup script for developers of this repo because we will need that test data to do any voting testing and this makes it a bit easier. We can secure this with a authentication using a dev token so that only we can hit it (will ticket for future implementation).

So after we have all the positions, we can create nominations. There is a new router and validations for it.

Also let me know what you think about this convention for validators:
1. In each router folder, we can create an enum called `LocalRoutes` and give a description of the route that we create.
2. When creating the validations using `express-validator`, we can use the naming convention `getValidations` that takes a `route: LocalRoutes` and returns the respective validations.
3. Call `getValidations` then `validate` on individual routes like usual.

The reason why I think we should refactor our route validation to follow that format is because at one point we ran into the issue where routes had the same suffix but not the entire route was the same so we had to get more specific with that param. I think if we use a shared enum (with a nice description) between the route and the validations, we won't run into that issue. It'll also be easier to tell which validations are for which route.

## Notes
Anything additional that reviewers should know going about your changes / going forward

### DIFF
--- a/src/routers/dev/index.ts
+++ b/src/routers/dev/index.ts
@@ -1,0 +1,34 @@
+import type { AuthenticatedRequest } from '../../types/network';
+
+import express, { Response, NextFunction } from 'express';
+import auth from '../../middleware/auth';
+import positions from '../../data/positions';
+import PositionModel from '../../models/position';
+
+const router = express.Router();
+
+// create positions if they don't exist
+router.post(
+  '/positions',
+  auth,
+  async (req: AuthenticatedRequest, res: Response, next: NextFunction) => {
+    try {
+      positions.forEach(async (p) => {
+        const existingPosition = await PositionModel.find({ title: p.title });
+        if (!existingPosition) {
+          const position = new PositionModel({
+            ...p,
+            isOpen: true,
+            occupant: req.user?._id,
+          });
+          await position.save();
+        }
+      });
+      res.send();
+    } catch (err) {
+      next(err);
+    }
+  },
+);
+
+export default router;
--- a/src/routers/dev/index.ts
+++ b/src/routers/dev/index.ts
@@ -13,17 +13,17 @@ router.post(
   auth,
   async (req: AuthenticatedRequest, res: Response, next: NextFunction) => {
     try {
-      positions.forEach(async (p) => {
-        const existingPosition = await PositionModel.find({ title: p.title });
-        if (!existingPosition) {
-          const position = new PositionModel({
-            ...p,
+      if (await PositionModel.find({}))
+        throw new Error('Positions already exist');
+      await PositionModel.create(
+        positions.map((position) => {
+          return {
+            ...position,
             isOpen: true,
             occupant: req.user?._id,
-          });
-          await position.save();
-        }
-      });
+          };
+        }),
+      );
       res.send();
     } catch (err) {
       next(err);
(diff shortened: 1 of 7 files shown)


src/routers/dev/index.ts
```diff
@@ -0,0 +1,34 @@
...
+            ...p,
+            isOpen: true,
+            occupant: req.user?._id,
+          });
+          await position.save();
+        }
+      });
+      res.send();
+    } catch (err) {
+      next(err);
```
jtaaa, line 29: Btw how do feel about [this `express-async-errors` package](https://github.com/davidbanham/express-async-errors). It would let us not have to catch just to send to next.

src/routers/nomination/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import validate from '../../middleware/validate';
+import NominationModel from '../../models/nomination';
+import getValidations from './validations';
+import { LocalRoutes } from './types';
+
+const router = express.Router();
+
+router.post(
+  '/',
+  getValidations(LocalRoutes.CREATE_NOMINATION),
```
jtaaa, line 14: I like this but I wish we could use `LocalRoutes` for the route path as well :( like we have for top-level routes.

If we swapped out the description for the route path and added the description in a comment we would still have the issue of POST, GET and all the others sharing the same path :'(

jtaaa:
I'm down for all of those conventions!

src/routers/dev/index.ts
```diff
@@ -0,0 +1,34 @@
...
+      positions.forEach(async (p) => {
+        const existingPosition = await PositionModel.find({ title: p.title });
+        if (!existingPosition) {
+          const position = new PositionModel({
+            ...p,
+            isOpen: true,
+            occupant: req.user?._id,
+          });
+          await position.save();
+        }
```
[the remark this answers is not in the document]
chriskrogh: Whaa. That's sick!
I might spend some time looking into `create`

src/routers/dev/index.ts
chriskrogh: I'm down for using it. If you want you can include it the next time that you're creating a route and we can ticket the refactor for all routes if it looks good.

src/routers/nomination/index.ts
chriskrogh: Actually that's why we want to use descriptions _instead_ of route paths. If we use descriptions, then regardless of whether the route was `POST`, `PATCH` etc, we would guarantee uniqueness.

What do you think?

src/routers/dev/index.ts
jtaaa: sg!

src/routers/nomination/index.ts
jtaaa: Yep I'm def for having each route maintain its validation. And using the description makes sense in that way too!

I just kinda wished we could have the route path information in the enum too somehow. but I dream 😢

src/routers/dev/index.ts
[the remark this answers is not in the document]
chriskrogh: updated in 32c74051bd90229c85ea941c95a00c1bbae05732

PR #80, opened 2020-09-03
Seconding a nominee

chriskrogh:
## What's changed
1. Updated the create nomination route to disallow duplicate nominations.
2. Created a route for seconding a nominee.
```
PATCH .../nomination/second?candidate={id}&position={id}
```
We pass a candidate id and a position id to find the nomination. Once we have the nomination, we can second it by adding our user id to the list of seconds. If we've already seconded the nomination before, then we get back an error.

## Notes
Danicia is asking on status updates so I feel like we should try to get some more progress on voting.

### DIFF
--- a/src/routers/nomination/index.ts
+++ b/src/routers/nomination/index.ts
@@ -16,11 +16,13 @@ router.post(
   auth(),
   async (req: AuthenticatedRequest, res: Response, next: NextFunction) => {
     try {
-      const existingNomination = await NominationModel.findOne({
-        position: req.body.position,
-        candidate: req.body.candidate,
-      });
-      if (existingNomination) throw new Error('Nomination already exists');
+      if (
+        await NominationModel.exists({
+          position: req.body.position,
+          candidate: req.body.candidate,
+        })
+      )
+        throw new Error('Nomination already exists');
       const nomination = new NominationModel({ ...req.body });
       await nomination.save();
       res.send();
--- a/src/routers/nomination/index.ts
+++ b/src/routers/nomination/index.ts
@@ -13,10 +13,41 @@ router.post(
   '/',
   getValidations(LocalRoutes.CREATE_NOMINATION),
   validate,
-  auth,
+  auth(),
   async (req: AuthenticatedRequest, res: Response, next: NextFunction) => {
     try {
+      const existingNomination = await NominationModel.findOne({
+        position: req.body.position,
+        candidate: req.body.candidate,
+      });
+      if (existingNomination) throw new Error('Nomination already exists');
       const nomination = new NominationModel({ ...req.body });
+      await nomination.save();
+      res.send();
+    } catch (err) {
+      next(err);
+    }
+  },
+);
+
+router.patch(
+  '/second',
+  getValidations(LocalRoutes.SECOND_NOMINEE),
+  validate,
+  auth(),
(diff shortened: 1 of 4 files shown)


src/routers/nomination/index.ts
```diff
@@ -13,9 +13,14 @@ router.post(
...
   validate,
-  auth,
+  auth(),
   async (req: AuthenticatedRequest, res: Response, next: NextFunction) => {
     try {
+      const existingNomination = await NominationModel.findOne({
+        position: req.body.position,
+        candidate: req.body.candidate,
+      });
+      if (existingNomination) throw new Error('Nomination already exists');
```
jtaaa, line 23: If we use `Model.count` here we avoid reading the doc from the DB.
https://stackoverflow.com/questions/27482806/check-if-id-exists-in-a-collection-with-mongoose

```suggestion
      const existingNominationCount = await NominationModel.count({
        position: req.body.position,
        candidate: req.body.candidate,
      });
      if (existingNominationCount > 0) throw new Error('Nomination already exists');
```

src/routers/nomination/index.ts
```diff
@@ -25,4 +30,30 @@ router.post(
...
+        position: req.query.position as string,
+        candidate: req.query.candidate as string,
+      });
+      if (!nomination) throw new Error('Could not find nomination');
+
+      const myId = req.user?._id;
+      if (nomination.seconds.includes(myId))
+        throw new Error('You already seconded this nomination');
+
+      nomination.seconds.push(myId);
```
jtaaa, line 50: Can we do `nominations.seconds.addToSet(myId);`? It doesn't always work in sys because of Typegoose complexities (I think bc it doesn't implicitly type arrays as Mongoose arrays) but it should work here. Then we don't even have to check if it already includes seconds!

jtaaa (dismissed):
LGTM!

Jeez only now reviewing this after 9 days. I'll try to show ACS some more love ❤️

src/routers/nomination/index.ts
jtaaa: Actually I think we can even do `Model.exists`

chriskrogh: Thanks for the recommendation! Updated here: c03449a194f061dee5ba595ae3e03f3c2fd66bf4

src/routers/nomination/index.ts
chriskrogh: I was taking a look at the https://mongoosejs.com/docs/schematypes.html and they don't have a `set` per say, but they do have a `map`. What do you think about using that instead of an `array`? What do you want to store the values as since the keys will be object ids.

jtaaa: I was actually just thinking to use this:
https://mongoosejs.com/docs/api.html#mongoosearray_MongooseArray-addToSet

MongoDB has the `$addToSet` update operator which is applied to an array and makes it act like a set

jtaaa: We may need to type the array differently in the schema to get access to the `addToSet` method. I'm not sure.

jtaaa (dismissed):

src/routers/nomination/index.ts
chriskrogh: updated in 037718410e26d9f8d1c9c24793ed8d96326eec43
#addToSet 🔥

jtaaa (approved):
Yessir!

PR #82, opened 2020-09-22
[⚓️x💯] Updated to Workspaces

chriskrogh:
## What's changed
I realized that it's not fun to work on acs because it feels a bit outdated compared to sys. So I updated our repo to use yarn workspaces so it has a similar look and feel.

I also updated `mern` and will probably deprecate `mern-auth` since passport is the 🌊.

## UI (if UI has changed)

Screen Shot 2020-09-22 at 12 15 08 AM

## Notes
Feels gr8!


packages/server/src/routers/dev/index.ts
```diff
@@ -13,7 +14,7 @@ router.post(
   auth(),
   async (req: AuthenticatedRequest, res: Response, next: NextFunction) => {
     try {
-      if ((await PositionModel.find({})).length === 0)
+      if ((await PositionModel.find({})).length !== 0)
```
chriskrogh, line 17: There might be an easier way for devs to add data to the db manually. I'm not a fan of this route (even though I created it) Might remove in a separate PR😅

tsconfig.json
```diff
@@ -1,13 +1,29 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES2015",
```
chriskrogh, line 3: Should we override this in the `client` config to be `es5`? Maybe for browser compatibility?

chriskrogh (commented):
After a fist pass on the code, things look okay. Mostly import resolution diffs

packages/server/src/routers/dev/index.ts
jtaaa: a script might work well for this

packages/server/src/types/network.ts
```diff
@@ -1,5 +1,5 @@
 import { Request } from 'express';
-import { UserDoc } from '../models/user';
+import { UserDoc } from '@acs/shared';
```
jtaaa, line 2:
```suggestion
import type { UserDoc } from '@acs/shared';
```

tsconfig.json
jtaaa: Yeah I think it doesn't hurt. Most browsers should be supporting es6 by now but IE still lives

jtaaa: I mean doesn't hurt to override it in the client! hurts a little to not support IE

jtaaa (commented):
The workspaces look clean af

Love that

How come you decided to move the models to the shared package? I find it's kinda strange that the schema is defined in shared.

chriskrogh:
> The workspaces look clean af
> 
> 
> 
> Love that
> 
> 
> 
> How come you decided to move the models to the shared package? I find it's kinda strange that the schema is defined in shared.

Just so that we can use the doc type in client. If we update the doc type, then I think we used to have to manually remember to update the client. Now we can just keep track of a 

ModelMetadata = {
  _id: ObjID;
  createdAt: Date;
  updatedAt: Date;
}

So we can do something like:

type User = UserDoc & ModelMetadata;

What do you think?

packages/server/src/routers/dev/index.ts
chriskrogh: I agree. I'll remove in a follow up PR and come up with a useful script

tsconfig.json
chriskrogh: I updated the base target to `es5` here 003a11077fdf545ffdc92455ac7df9fdfb3ef6c1 so that all our packages transpile to the same version of js

jtaaa:
Yeah I agree. What about importing the `UserDoc` from the server?

chriskrogh:
@jtaaa b72317420f8666e96d88570b4b394761bc49a42a
Moved docs to `shared`, schemas in `server`

packages/shared/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "@acs/shared",
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "license": "MIT",
+  "scripts": {
+    "build": "tsc"
+  },
+  "dependencies": {
+    "mongoose": "^5.10.6"
```
jtaaa, line 11: I wonder if it would be possible to move `mongoose` to dev deps. I think we're only using it for types

packages/client/.storybook/main.js
```diff
@@ -0,0 +1,11 @@
...
+  "stories": [
+    "../src/**/*.stories.mdx",
+    "../src/**/*.stories.@(js|jsx|ts|tsx)"
+  ],
+  "addons": [
+    "@storybook/addon-links",
+    "@storybook/addon-essentials",
+    "@storybook/preset-create-react-app"
+  ]
+}
```
jtaaa, line 11: Missing some trailing newlines on these `.js` storybook files
```suggestion
}

```

packages/client/.storybook/preview.js
```diff
@@ -0,0 +1,4 @@
+
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+}
```
jtaaa, line 4:
```suggestion
}

```

jtaaa (approved):
LGTM 💯

packages/shared/package.json
chriskrogh: sgtm. Should be fine. updated in 26a090b84e7934dd9f981a867fe3ff89d09bf9d3

PR #84, opened 2020-10-10
⭕️ CI

chriskrogh:
Added the boy ⭕️
This is a cool high level reference sheet with examples for [circle](https://circleci.com/docs/reference-2-1/#section=configuration) 
Also their docs are pretty amazing. I'm a big fan.

Since you're a code owner, you should have access to the app here: https://app.circleci.com/pipelines/github/uwaterlooacs

### Builds
When you push to any branch, we run a few build jobs. If we're lucky they might look like this:

Screen Shot 2020-10-10 at 5 39 57 PM

We build `shared`, then store its build output in circle's cache so that the build jobs for `client` and `server` can use it.
Each job runs a fresh build (installing its own dependencies on a fresh instance)

Once the branch build passes, and your branch is merged with `master`, `master` runs a new build workflow to see if the merged code works, and then runs `deploy`.

### Deploys
I set up our EC2 instance with a new ssh key that circle will use to connect. If the build on master is successful, then we run the deploy job. This just `ssh` s into our EC2 and runs the deploy script.

This is slightly lame. If you think about it, we just built the new code with master so we know that it works. An alternative was to `scp` the build output from the circle jobs over to the EC2, but that's a lot of files and an even worse solution imo. 
So we have 1 redundant build, but I think it's necessary for now.

Furthermore, in the future it would be pretty cool to use a service like ECS to use containers in our deploys. We could also write our own custom Docker image (and have circle use that).

## Notes
The [circle cli](https://circleci.com/docs/2.0/local-cli/) is actually pretty cool. You can run jobs locally and test to see whether the config is valid.

Circle also has the concept of [orbs](https://circleci.com/docs/2.0/orb-intro/) which are popular configs that people use that we can just import into our config (importing one YAML file into ours). So we could write nice deploy code and not just ssh. Would be cool to use on other projects maybe.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,69 @@
+version: 2.1
+
+executors:
+  node:
+    docker:
+      - image: circleci/node:lts
+  base:
+    docker:
+      - image: cimg/base:2020.01
+
+commands:
+  init:
+    steps:
+      - checkout:
+          path: ~/project
+      - run: yarn --frozen-lockfile
+
+jobs:
+  shared:
+    executor: node
+    working_directory: ~/project/packages/shared
+    steps:
+      - init
+      - run: yarn ci:build
+      - save_cache:
+          key: shared-cache
+          paths:
+            - ~/project/packages/shared/dist
+  client:
+    executor: node
+    working_directory: ~/project/packages/client
+    steps:
+      - init
+      - restore_cache:
+          keys:
+            - shared-cache
+      - run: yarn ci:build
+  server:
+    executor: node
+    working_directory: ~/project/packages/server
+    steps:
+      - init
+      - restore_cache:
+          keys:
+            - shared-cache
+      - run: yarn ci:build
+  deploy:
+    executor: base
+    steps:
+      - run: ssh -o StrictHostKeyChecking=no ec2-user@ec2-18-220-45-225.us-east-2.compute.amazonaws.com "~/acs/scripts.deploy.sh"
+
+workflows:
+  build:
+    jobs:
+      - shared
+      - client:
+          requires:
+            - shared
+      - server:
+          requires:
+            - shared
+      - deploy:
+          filters:
+            branches:
+              only:
+                - master
+          requires:
+            - client
+            - server
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -5,6 +5,7 @@
   "license": "MIT",
   "scripts": {
     "build": "tsc",
+    "ci:build": "NODE_ENV=test node ../../node_modules/typescript/lib/tsc",
     "dev": "ts-node-dev --inspect --respawn --transpile-only src/index.ts"
   },
   "dependencies": {
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -5,7 +5,8 @@
   "typings": "dist/index.d.ts",
   "license": "MIT",
   "scripts": {
-    "build": "tsc"
+    "build": "tsc",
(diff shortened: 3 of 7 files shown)


chriskrogh:
Hey Josh,
I'm just going to merge without approval because I want to test the deploy job. 
Please review as normal! Will follow up any feedback in a new PR

packages/server/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "MIT",
   "scripts": {
     "build": "tsc",
+    "ci:build": "NODE_ENV=test node ../../node_modules/typescript/lib/tsc",
```
jtaaa, line 8: why can't we just use `tsc`?

packages/shared/package.json
```diff
@@ -5,7 +5,8 @@
   "typings": "dist/index.d.ts",
   "license": "MIT",
   "scripts": {
-    "build": "tsc"
+    "build": "tsc",
+    "ci:build": "NODE_ENV=test node ../../node_modules/typescript/lib/tsc"
```
jtaaa, line 9: same here why not `NODE_ENV=test tsc`?

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+    steps:
+      - init
+      - restore_cache:
+          keys:
+            - shared-cache
+      - run: yarn ci:build
+  deploy:
+    executor: base
+    steps:
+      - run: ssh -o StrictHostKeyChecking=no ec2-user@ec2-18-220-45-225.us-east-2.compute.amazonaws.com "~/acs/scripts.deploy.sh"
```
jtaaa, line 50: do you set the ssh key in Travis? 😮

jtaaa (commented):
Looks cleaner and more easily understandable than Travis imo

packages/server/package.json
chriskrogh: We need typescript on our docker image to build the code but after running `yarn`, we'll have a copy of this package already installed using the version of ts that we use in the proj. So instead of doing something like 

```yarn add global typescript```

we just use the tsc that we already have.

I actually tried to `yarn add global typescript` as one of the commands to run in the init stage of each job, but that won't work because the job doesn't have the right permissions to modify the docker container directly.

The other alternative would be to write our own `Dockerfile` and [tell circle to use that](https://circleci.com/docs/2.0/custom-images/). I kind of didn't want to do that because it seemed a bit unnecessary just for tsc. But it would be nice to have! Feel free to make a PR if you want!

.circleci/config.yml
chriskrogh: yessir in ⭕️

chriskrogh:
@jtaaa Just as an update, the pipeline fully works as described in the PR description! The deploy script needed an additional line to set the working directory, but everything works! Automatic deploys once things are merged!

Screen Shot 2020-10-12 at 9 56 29 AM

packages/server/package.json
jtaaa: Hmm but we should be able to reference the local tsc install the same way as we do in the build script. Instead of using the relative path to it in node_modules

jtaaa: I don't think it needs to be installed globally. I'm wondering why we need to use the relative path in this case
https://docs.npmjs.com/misc/scripts#path

chriskrogh: Maybe the error I got had something to do with the monorepo setup. The TS package is hoisted to root, so maybe that's why? I have an idea of what we can try. Will follow up with a PR

PR #85, opened 2020-10-12
Updated ⭕️ build scripts

chriskrogh:
## What's changed
We now just call `tsc` instead of `node ../../node_modules/blahblah/tsc`

## Notes
⭕️ builds


jtaaa (approved):
🚀

PR #92, opened 2021-01-03
Finalize position winners when voting is complete

jtaaa:
## What's changed

Update the PositionDocs with new occupant and `isOpen = false` when the voting stage is updated through PATCH at `/voting/stage` from the vote stage to the results stage.


chriskrogh (approved):

chriskrogh (changes requested):
.

chriskrogh (approved):

PR #93, opened 2021-01-03
Decline Nomination

chriskrogh:
## What's changed
Added route for declining a nomination


jtaaa (approved):
LGTM.
Starting to wonder if it makes more sense for the `/second` and `/decline` routes to just take the nomination's `_id` instead of `position` and `candidate`.

PR #97, opened 2021-01-04
Get Stage Route

chriskrogh:
## What's changed
Gets current voting stage


packages/server/src/routers/voting/index.ts
```diff
@@ -11,6 +11,19 @@ import { finalizeResults } from './utils';
 
 const router = express.Router();
 
+// get current voting stage
+router.get(
+  '/stage',
+  auth({ isAdmin: true }),
```
jtaaa, line 17:
```suggestion
  auth(),
```

jtaaa (approved):

